Issue #62, opened 2021-12-10
HTTPS Support


Weldawadyathink:
HTTPS Support

https://tailscale.com/kb/1153/enabling-https/

Is it possible to add support for requesting an SSL certificate with Tailscale?

franzbischoff:
There is a requirement on current version: https://issueexplorer.com/issue/tailscale/tailscale/2932

Fix: 
create a folder at /data/tailscale
edit `/etc/services.d/tailscaled/run`
change `options+=(--state=/data/tailscaled.state)` to `options+=(--state=/data/tailscale/tailscaled.state)`
copy the current `/data/tailscaled.state` to `/data/tailscale/tailscaled.state` to keep login etc.
restart service: `s6-svc -r /var/run/s6/services/tailscaled`

now you can run `/opt/tailscale cert yourmachine.yourdomain.ts.net`

benley:
> There is a requirement on current version: https://issueexplorer.com/issue/tailscale/tailscale/2932

I get a 404 on that link, but I think this is the issue in question: https://github.com/tailscale/tailscale/issues/2932

franzbischoff:
Just made a Gist for that.

Use at your own risk:

[tailscale hack](https://gist.github.com/franzbischoff/1dc960aa5ad93c604f1cc57282f3ebb4)

ananthb:
Would it make sense to add this as a config option?

franzbischoff:
For me makes total sense. To be able to move from using an external port on my router to access HA with my phone, for example, I use HTTPS and the tailscale "domain" with the provided certificate.

ananthb:
@frenck could you take a look at my PR? Would love to get this merged soon!

benley:
not stale

ananthb:
Still being worked on.

benley:
still not stale

franzbischoff:
not stale!

martusi61:
I'm waiting too!

franzbischoff:
nooooooooooot stale

ceiphr:
not stale

stroodle96:
Also hoping this gets added.

Mincka:
You can have a look at this alternative addon which seems to provide https support:
https://github.com/tsujamin/hass-addons/tree/main/tailscale
https://github.com/tsujamin/hass-addons/blob/main/tailscale/DOCS.md#option-cert_domain

elcajon:
[...]

PR #89, opened 2022-04-09
Request TLS certificate for tailnet domain alias.

ananthb:
# Proposed Changes

Request TLS certificate from tailscale. HTTPS must be enabled for the tailnet.
https://tailscale.com/kb/1153/enabling-https/

## Related Issues

https://github.com/hassio-addons/addon-tailscale/issues/62

@lmagyar has an even easier solution in #137.

### DIFF
--- a/tailscale/config.yaml
+++ b/tailscale/config.yaml
@@ -19,9 +19,6 @@ arch:
 init: false
 map:
   - ssl:rw
-options:
-  certfile: fullchain.pem
-  keyfile: privkey.pem
 hassio_api: true
 privileged:
   - NET_ADMIN
@@ -31,5 +28,3 @@ host_network: true
 schema:
   tags: ["match(^tag:[a-zA-Z0-9]-?[a-zA-Z0-9]+$)?"]
   domain_alias: match(^[a-z0-9][a-z0-9\-_]{0,61}\.ts\.net$)?
-  certfile: str
-  keyfile: str
--- a/tailscale/config.yaml
+++ b/tailscale/config.yaml
@@ -20,7 +20,6 @@ init: false
 map:
   - ssl:rw
 options:
-  request_cert: false
   certfile: fullchain.pem
   keyfile: privkey.pem
 hassio_api: true
@@ -31,7 +30,6 @@ devices:
 host_network: true
 schema:
   tags: ["match(^tag:[a-zA-Z0-9]-?[a-zA-Z0-9]+$)?"]
-  request_cert: bool
   domain_alias: match(^[a-z0-9][a-z0-9\-_]{0,61}\.ts\.net$)?
   certfile: str
   keyfile: str
--- a/tailscale/config.yaml
+++ b/tailscale/config.yaml
@@ -17,6 +17,12 @@ arch:
   - armv7
   - i386
 init: false
+map:
+  - ssl:rw
+options:
+  request_cert: false
+  certfile: fullchain.pem
+  keyfile: privkey.pem
 hassio_api: true
 privileged:
   - NET_ADMIN
@@ -25,3 +31,7 @@ devices:
 host_network: true
 schema:
   tags: ["match(^tag:[a-zA-Z0-9]-?[a-zA-Z0-9]+$)?"]
+  request_cert: bool
+  domain_alias: match(^[a-z0-9][a-z0-9\-_]{0,61}\.ts\.net$)?
+  certfile: str
+  keyfile: str
--- a/tailscale/rootfs/etc/services.d/cert/run
+++ b/tailscale/rootfs/etc/services.d/cert/run
@@ -8,8 +8,8 @@
 tailscale_cert() {
   if bashio::config.has_value 'domain_alias'; then
     /opt/tailscale cert \
-      --cert-file "/ssl/$(bashio::config 'certfile')" \
-      --key-file "/ssl/$(bashio::config 'keyfile')" \
+      --cert-file "/ssl/tailscale/fullchain.pem" \
+      --key-file "/ssl/tailscale/privkey.pem" \
       "$(bashio::info.hostname).$(bashio::config 'domain_alias')"
   fi
 }
--- a/tailscale/rootfs/etc/services.d/cert/run
+++ b/tailscale/rootfs/etc/services.d/cert/run
@@ -4,10 +4,9 @@
 # Runs tailscale cert periodically
 # ==============================================================================
 
-# Request TLS certificate if configured
+# Request TLS certificate if domain_alias is set
 tailscale_cert() {
-  if bashio::config.true 'request_cert'; then
-    bashio::config.require 'domain_alias'
+  if bashio::config.has_value 'domain_alias'; then
     /opt/tailscale cert \
       --cert-file "/ssl/$(bashio::config 'certfile')" \
       --key-file "/ssl/$(bashio::config 'keyfile')" \
@@ -15,7 +14,7 @@ tailscale_cert() {
   fi
 }
 
-cert_err='Error running `tailscale cert`. Visit https://tailscale.com/kb/1153/enabling-https/ to setup HTTPS.'
+cert_err='Error running "tailscale cert". Visit https://tailscale.com/kb/1153/enabling-https/ to setup HTTPS.'
 
 if ! tailscale_cert; then
   bashio::exit.nok "$cert_err"
--- a/tailscale/rootfs/etc/services.d/cert/run
+++ b/tailscale/rootfs/etc/services.d/cert/run
@@ -1,0 +1,31 @@
+#!/usr/bin/with-contenv bashio
+# ==============================================================================
+# Home Assistant Community Add-on: Tailscale
+# Runs tailscale cert periodically
+# ==============================================================================
+
+# Request TLS certificate if configured
+tailscale_cert() {
+  if bashio::config.true 'request_cert'; then
+    bashio::config.require 'domain_alias'
+    /opt/tailscale cert \
+      --cert-file "/ssl/$(bashio::config 'certfile')" \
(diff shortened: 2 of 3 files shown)


ananthb:
`tailscale cert` has to be run periodically to keep certs fresh. I can either run a background script from the post script, or start a cron daemon and use that. Any suggestions here?

franzbischoff (commented):
Right!
1. add ssl folder to the addon
2. add options
3. fetch new cert every day

👍

ananthb:
Yep that's what it does. The cert process runs once every 24 hours to fetch a new cert. The SSL folder is mapped rw. There is an option to enable the addon and a mandatory one for the tailnet domain name.

franzbischoff:
The PR is failing because of the target name :-/

frenck:
> The PR is failing because of the target name :-/

Its not.

ananthb:
Yes automation seems to expect a label but the PR is currently unlabeled. I don't have permission to label it either.

frenck:
Yup, that is not a issue

ananthb:
@frenck so how do we proceed?

frenck:
By having patience

ananthb:
@frenck is this failure unexpected? I don't understand the error message.

frenck:
>  is this failure unexpected?

Yes

> I don't understand the error message.

OK... not sure what to do about that 🤷 It complains a label is missing, so that needs to be added by someone with permissions to do that.

ananthb:
I meant after you added the label, the amd64 build failed at a docker buildx step. I was wondering if i had to fix something but that doesn't appear to be the case.

tailscale/config.yaml
```diff
@@ -17,6 +17,12 @@ arch:
   - armv7
   - i386
 init: false
+map:
+  - ssl:rw
+options:
+  request_cert: false
+  certfile: fullchain.pem
+  keyfile: privkey.pem
```
frenck, line 25: Maybe we can just default these and remove the config?

For example by storing them in `/ssl/tailscale/[fullchain|privkey].pem`?

tailscale/config.yaml
```diff
@@ -25,3 +31,7 @@ devices:
 host_network: true
 schema:
   tags: ["match(^tag:[a-zA-Z0-9]-?[a-zA-Z0-9]+$)?"]
+  request_cert: bool
+  domain_alias: match(^[a-z0-9][a-z0-9\-_]{0,61}\.ts\.net$)?
```
frenck, line 35: Maybe we don't need `request_cert` at all? For example, if one provides the domain alias, that should already be enough?

tailscale/rootfs/etc/services.d/cert/run
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/with-contenv bashio
+# ==============================================================================
+# Home Assistant Community Add-on: Tailscale
+# Runs tailscale cert periodically
+# ==============================================================================
+
+# Request TLS certificate if configured
+tailscale_cert() {
+  if bashio::config.true 'request_cert'; then
+    bashio::config.require 'domain_alias'
```
frenck, line 10: we don't have to check this constantly (every 24 hours). If this is kept, we can check it once (in a `cont-init` script).

frenck (changes requested):

tailscale/config.yaml
franzbischoff: I'm using as @frenck says. IMHO it is safer to have a different folder for tailscale as you may still need the non-tailscale certificates for something else. You can define which TSL the services will use, by-service, instead of forcing all [working] services to use tailscale's and for some reason all breaks.

An use case I can provide is the inability to login to HA itself just because I was relying on tailscale, and the tailscale addon went down.

my 2 cents

tailscale/config.yaml
franzbischoff: And if we want to temporarily stop renewal?

tailscale/rootfs/etc/services.d/cert/run
franzbischoff: Can we at least auto-check every week? Or, grab the expire date and schedule it in someway. I know we frequently have updates and restart all services (and TS would try to renew the certificates), but some users or at some period we could have it untouched for a long time.

Anyway, if the certificate expires the option for the user is to restart the container, right?

tailscale/config.yaml
frenck: That is a shallow use case to justify it imho. If wanted, one could temporary remove the alias as well 🤷

tailscale/config.yaml
ananthb: I'd like the option to choose, because I'm going to run HTTPS only on the tailscale address and firewall everything else. We could change the default to something else but i went with what other certificate add-ons are doing.

tailscale/config.yaml
ananthb: Yeah makes sense. We can do without request_cert.

tailscale/rootfs/etc/services.d/cert/run
ananthb: Hmm. The issue is renewal. If i have a long uptime without updates or restarts i still want new certs to be fetched automatically. The bash sleep is super light weight so i went with it.

tailscale/config.yaml
franzbischoff: @frenck yes, it is a shallow use case, I'm fine with removing the `request_cert`. My "concern" was that if someone uses a complex regex, he would have to save it somewhere. But anyway, still shallow :-)

ananthb: Users don't input regex though. The regex validates the config option which expects a domain name in the form xyz.ts.net where xyz is your tailnet domain alias from the tailscale admin.

tailscale/config.yaml
frenck: > I'd like the option to choose

I'd like to reduce complexity and remove it.

ananthb: Fair enough. I'll remove it.

franzbischoff:
Not staled!

martusi61:
I'm waiting too!

nsiicm0:
@frenck I know it is not on us to accelerate the process, but could you have a look again at this PR? AFAIK, it has been sitting now for 3 months and all the changes you requested have been implemented. This feature would be a real game changer for the addon.

Thank you so much!

frenck:
I am abroad for a couple of weeks, so can't do anything now. Marked it to look at when I get back.

ShadowJonathan:
(Not stale)

ShadowJonathan:
(Still not stale)

ananthb:
@frenck did you have any time to look at this? I'm starting to feel that it's taking so long because it directly competes with hassio's paid offering.

If you're not interested in this change, I'm more than happy to maintain a more up to date fork.

ShadowJonathan:
(Not stale)

ananthb:
@frenck this change is eagerly awaited by me and quite a few others it would seem. Any chance it gets merged soon?

If you don't want this in change please let us know. It's pretty inconsiderate to just leave us hanging.

ShadowJonathan:
Anecdotally, I'm also awaiting this change as I've moved over every homelab-hosted service to tailscale HTTPS a few months ago, save for Home Assistant, because of this PR's timeline, currently I'm still using cloudflare and a port-forward instead.

ananthb:
For what it's worth, I've stopped using homeassistant entirely because of the petty handling of this change. I will keep this PR open until they decide to do something about it.

lmagyar:
Until it is merged I forked the repo, updated it to Tailscale v1.34.2, and bumped the add-on base image also to v13.0.1.

And I merged this PR and made some modifications and bugfixes:
- the Configuration option is `cert_domain` (not `domain_alias`), it is much more logical, it is analog to the `tailscale cert <domain>` command
- it creates the `/ssl/tailscale` directory before fetching the cert
- uses `--statedir=/data` instead of `--state=/data/tailscaled.state`, see https://github.com/tailscale/tailscale-qpkg/pull/64
- fetch certificate on start and only within 1 week before expiration
- modified the daemon finish script to the usual script, I know new add-ons have better, it can be bumped later with other scripts

See: https://github.com/lmagyar/homeassistant-addon-tailscale

Disclaimer: I'm just a tinkerer, and "it works on my machine..." (tested on armv7)

I will create some new PR-s soon.

franzbischoff:
Hi

My domain is from the beginning and doesnt comply with your regex filter...
I use only letters and dashes for ex, and the regex fails...
--
Francisco Bischoff


Laszlo Magyar ***@***.***> escreveu no dia sábado, 7/01/2023
à(s) 21:38:

lmagyar:
> Hi My domain is from the beginning and doesnt comply with your regex filter...
> I use only letters and dashes for ex, and the regex fails...

Could you please send what is in your config? `cert_domain: ?????`

Maybe you entered the machine name also, like `cert_domain: abcd.tail1234.ts.net`, instead of only `cert_domain: tail1234.ts.net`

franzbischoff:
I found that what you ask is the main domain: xyz.ts.net
I was trying the device subdomain: myha.xyz.ts.net

it works as it was :)
sorry

sirreal:
I'm still hoping this progresses. Not stale.

lmagyar:
I think #137 is a better solution, because that doesn't require any configuration and also provides a built-in proxy by tailscaled.

ananthb:
@lmagyar that does look better!

I'm not holding my breath waiting for either one to get merged though. The response from Nabu Casa is pretty clear now.

> By having patience

@frenck this patient enough for you, or should we continue waiting forever?

Honestly this single issue has really turned me off of homeassistant. I'm a big believer in the spirit behind open source and Nabu Casa has shown absolutely none of that here. The community should really question their commitment to all of the projects they control.

PR #104, opened 2022-07-04
Add an option for accepting DNS.  Fixes #74

### LINKED ISSUE #74: Magic DNS, HA & AdGuard/Pi-Hole
According to https://tailscale.com/kb/1114/pi-hole/

> When running tailscale up, you’ll need to pass the `--accept-dns=false` flag. Pi-Hole uses DNS servers configured within Linux as its upstream servers, where it will send DNS queries that it cannot answer on its own. Since we’re going to make the Pi-Hole be our DNS server, we don’t want Pi-Hole trying to use itself as its own upstream.

It would be nice to disable DNS as an option: `--accept-dns=false` so we can use HA - AdGuard as the DNS server for all tailscale nodes.

https://github.com/hassio-addons/addon-tailscale/blob/cb67a29a25c4dc71648ef01b29b806e25a6a2351/tailscale/rootfs/etc/services.d/tailscaled/post#L53-L59

chrismgrayftsinc:
# Proposed Changes

Add an option to allow tailscale not to accept DNS, allowing the HomeAssistant node to be used as a DNS server.

## Related Issues

#74

### DIFF
--- a/tailscale/Dockerfile
+++ b/tailscale/Dockerfile
@@ -13,7 +13,7 @@ RUN \
         ipcalc=1.0.1-r0 \
         iproute2=5.17.0-r0 \
         iptables=1.8.8-r1 \
-        nginx=1.22.0-r0 \
+        nginx=1.22.0-r1 \
     \
     && ln -sf /sbin/xtables-nft-multi /sbin/ip6tables \
     && ln -sf /sbin/xtables-nft-multi /sbin/iptables \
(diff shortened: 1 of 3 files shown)


chrismgrayftsinc:
I have now tried this locally and have found that the adguard extension can't bind to the tailscale interface (because the tailscale is in a container).  So it's a bit pointless for my use case.

Issue #96, opened 2022-05-18
Only one of two subnets is advertised


andorardo:
# Problem/Motivation

I have two subnets on my HA host, on two NICs. `ha net info` shows as much.

They are 192.168.X.0/24 and 192.168.Y.0/24

## Expected behavior

In tailscale, that both subnets appear under "Subnet routes"

## Actual behavior

only the first subnet shows up under Subnet routes

## Steps to reproduce

Use two subnets on two NICs?

## Proposed changes

Either a way to detect both subnets, or a way to configure the `advertise-routes` option

andorardo:
Please don't close this as "completed", since it is still very much an issue!

dalkain:
Also interested in this being implemented

SoulEater45:
It would be nice to have the possibility to enter further subnets in the config as an array. If I have time in the upcoming days, maybe I can try to configure it, since it should be just an array joined with a comma and slapped onto the `--advertise-routers` flag.

DirschedlF:
I'm still interested to specify more than one subnet. See my comment regarding "--advertise-routes= ..." above.

ocsi01:
I am also interested in this to be solved. Having both LAN and WIFI on my Hass devices, it would be nice to have both as shareable subnets.

Mincka:
In the past, I used this alternative addon for tailscale, when this one was broken.
https://github.com/tsujamin/hass-addons/tree/main/tailscale

Since I need multiple route support now and there's no support for it, I'm switching back to @tsujamin's addon.
It offers a more recent version and a lot more possibilities regarding the options, especially route advertising.

My config looks like this in YAML and multiple routes support is working great.

```
hostname: homeassistant
auth_key: tskey-auth-xxx
advertise_routes: 10.0.0.0/24,192.168.0.0/24
userspace_networking: true
advertise_exit_node: true
accept_routes: true
```

dalkain:
Still needed

novirium:
PR #116 _does_ work, and I'm using an almost identical modification to be able to route both of my ethernet interfaces via Tailscale.  I'm not sure why it was closed - I'd be happy to submit another PR if that's all that's needed to get this fix in?

lmagyar:
There is a working PR #135 that solves this issue: advertise all supported interfaces as Tailscale Subnets not only the default interface.

DezzyMelb:
[...]

PR #116, opened 2022-10-03
Add support for additional interfaces

elcajon:
# Proposed Changes

Get interfaces and related ip addresses instead of using main interface only.
[Coding reused](https://github.com/hassio-addons/addon-adguard-home/blob/8a5af900ca46ff480c174882361a03f3fa2517e1/adguard/rootfs/etc/cont-init.d/adguard.sh#L52-L56) from AdGuardHome add-on.

## Related Issues

#96 

[autolink-references]: https://help.github.com/articles/autolinked-references-and-urls/

### DIFF
--- a/tailscale/rootfs/etc/services.d/tailscaled/post
+++ b/tailscale/rootfs/etc/services.d/tailscaled/post
@@ -3,14 +3,18 @@
 # Home Assistant Community Add-on: Tailscale
 # Runs after the machine has been logged in into the Tailscale network
 # ==============================================================================
+declare -a interfaces
 declare -a addresses
 declare -a routes
 declare ipinfo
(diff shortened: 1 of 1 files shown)


ocsi01:
I installed this add-on including this PR to my HASS setup as a local add-on.
It does not share any of my NIC's subnet. ( Compared to the released 0.7 version, which shares my first NIC.)

elcajon:
I have successfully connected two tagged VLANs via Tailscale with the configuration (HA OS) for quite some time. Can you share some information about your system environment?

ocsi01:
Not sure why, but after restarting the add-on, now it works. I can see both of my subnets shared. Thanks!

PR #138, opened 2023-01-18
Suppress tailscaled logs after 200 lines

lmagyar:
# Proposed Changes

Tailscaled is quite chatty and emits a really large amount of log messages.

This PR disables these messages after 200 lines, but makes it possible to re-enable the full logging for troubleshooting.

Based on the idea from https://github.com/tailscale/tailscale-synology/pull/110

## Related Issues

--

### DIFF
--- a/tailscale/rootfs/etc/s6-overlay/s6-rc.d/tailscaled/run
+++ b/tailscale/rootfs/etc/s6-overlay/s6-rc.d/tailscaled/run
@@ -9,7 +9,7 @@ declare -a options
 bashio::log.info 'Starting Tailscale...'
 
 options+=(--state=/data/tailscaled.state)
-# Opt out of client log upload
+# Opt out of client log upload to log.tailscale.io
 options+=(--no-logs-no-support)
 options+=(--tun=userspace-networking)
 
@@ -18,7 +18,9 @@ options+=(--tun=userspace-networking)
 
 # Run Tailscale
 if bashio::debug ; then
-    exec /opt/tailscaled "${options[@]}"
+  exec /opt/tailscaled "${options[@]}"
 else
-    exec /opt/tailscaled "${options[@]}" &> /dev/null
+  bashio::log.notice "Tailscale logs will be suppressed after 200 lines, set add-on's configuration option 'log_level' to 'debug' to see further logs"
+  set -o pipefail
+  /opt/tailscaled "${options[@]}" 2>&1 | stdbuf -i0 -oL -eL sed -n -e '1,200p' -e "201c[further tailscaled logs suppressed, set add-on's configuration option 'log_level' to 'debug' to see further tailscaled logs]" 
 fi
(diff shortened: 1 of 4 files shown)


LICENSE.md
```diff
@@ -1,6 +1,6 @@
 # MIT License
 
-Copyright (c) 2021-2022 Franck Nijhof
+Copyright (c) 2021-2023 Franck Nijhof
```
frenck, line 3: Unrelated changes

frenck (changes requested):
> Tailscaled is quite chatty and emits a really large amount of log messages.


I'm missing: Why is that a concern?

lmagyar:
> > Tailscaled is quite chatty and emits a really large amount of log messages.
> 
> I'm missing: Why is that a concern?

Tailscaled writes into the log something in nearly each second.
- This debug noise hides any useful log entry.
- The logs finally are written to SD card, so I prefer to minimize the wear.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/tailscaled/run
```diff
@@ -9,10 +9,16 @@ declare -a options
 bashio::log.info 'Starting Tailscale...'
 
 options+=(--state=/data/tailscaled.state)
+# Opt out of client log upload
+options+=(--no-logs-no-support)
```
frenck, line 13: 👍

tailscale/rootfs/etc/s6-overlay/s6-rc.d/tailscaled/run
```diff
@@ -9,10 +9,16 @@ declare -a options
...
 
 # Trigger post process to run, after stuff got connected
 ./post &
 
 # Run Tailscale
-exec /opt/tailscaled "${options[@]}"
+if bashio::debug ; then
+    exec /opt/tailscaled "${options[@]}"
+else
+    exec /opt/tailscaled "${options[@]}" &> /dev/null
```
frenck, line 23: I'm not a fan of this, mostly because it isn't controlling log levels; it is just suppressing all output.

The problem with that is that a user ends up with 'nothing', making them wonder if it works even (we get lots of complains with other add-ons that actually use software that don't output anything by nature).

The tailscale project has been reducing logs output quite a lot over time. As for writing logs, I don't consider that a concern tbh; if we can reduce the log level, sure! I'll be all for it, but that isn't the case here, as we already run on log level 0 (the default).

../Frenck

frenck (changes requested):

lmagyar:
I've spent a few hours googling, and found the same discussion in tailscaled-synology: https://github.com/tailscale/tailscale-synology/pull/110

Solution: the logs suppressed after 200 lines, so the startup can be seen in details, but after a few minutes it will stop emitting noise.

I had to modify it a bit due to busybox/sed limitations.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/tailscaled/run
```diff
@@ -9,10 +9,18 @@ declare -a options
...
 # Trigger post process to run, after stuff got connected
 ./post &
 
 # Run Tailscale
-exec /opt/tailscaled "${options[@]}"
+if bashio::debug ; then
+  exec /opt/tailscaled "${options[@]}"
+else
+  bashio::log.notice "Tailscale logs will be suppressed after 200 lines, set add-on's configuration option 'log_level' to 'debug' to see further logs"
+  set -o pipefail
```
frenck, line 24: That already defaults in our bashio environment.

```suggestion
```

frenck (changes requested):

frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

Issue #169, opened 2023-04-02
Routes not added in HA


CamFlyerCH:
# Problem/Motivation

To start: I am not entirely sure, if this is an issue, a technical limitation or a feature request. I am in the situation, where I have sensors in 2 separate networks / locations that I want to connect to from one HA instance.

## Expected behavior

I have HA in network 1 and HA also in network 1. This is all fine.
I have sensors in location 2 and also a running tailscale node that act also as a subnet router. Sensors are reachable with their local IP from any tailscale node in the same tailscale network.
I install also tailscale via Add-on on HA in network 1, add it to the same tailscale network. So I can include now sensors from network 2 via tailscale VPN to the HA in network one by using their local IP.

## Actual behavior

Mostly all work as expected till the last part. The sensors in network two are reachable from any tailscale node, but not from HA in network 1.

## Steps to reproduce

I tested it with Integrations like SolarLog or with ping from a HA CLI . Also I connected to the console of the tailscale Add-on. Same issues and also there are no routes visible with using "ip route" or "route -n".

## Proposed changes

Get routing working. I saw that "tailscale up"  contains the needed option "--accept-routes". But it seams not to work. Also found no error in the log (debug - mode).
(Would be nice if also https://tailscale.com/kb/1214/site-to-site/ would be possible.)

CamFlyerCH:
Thanks for the feedback. First I am wondering, if it would be possible to change the add-on to not run in userspace. I checked the Wireguard add-on, and it has an device wg0. I would understand, but don't know how, if it would be needed to add some static routes, but how is this even possible in a hassio setup ?

The other question, where I did not found hints for, ist how I could use the SOCKS or HTTP proxy to access just some sensors ? For sensors attached "the other way" like with MQTT I understand the issue lies on the network 2 to be able to route the traffic to the HA. But for that direction (lets say incoming to HA) the tailscale add-on is already working fine.

frenck:
With the more recent OS releases, we could consider moving out of the user space.

Issue #129, opened 2022-12-31
Cannot access tailscale network devices from HA


george-viaud:
# Problem/Motivation
I wish to connect to other tailscale devices that host services from within Home Assistant

## Expected behavior
When addon-talescale is installed, other tailscale connected devices are reachable from HA

## Actual behavior
Unable to reach devices (even though those devices can reach the HA client)

Example from within HA terminal where x.x.x.x is the IP of a tailscale-connected device that has simple webserver running: 
```
curl http://x.x.x.x
cURL error 28: Connection timed out after X milliseconds
```
## Steps to reproduce
Connect HA to tailscale and confirm you can reach the HA instance from another tailscale client

Setup some kind of simple service to test with (http in my case) and confirm this is reachable as well from another tailscale client

Attempt to connect to the test service from the HA instance using the same IP that worked in step 2

lmagyar:
As I understand the add-on is running in userspace networking (`options+=(--tun=userspace-networking)` is hard wired), so it doesn't create a network interface that's visible from other containers, so its not possible to route traffic (see https://tailscale.com/kb/1112/userspace-networking/).

Try: https://github.com/tsujamin/hass-addons

Currently I'm testing the different Tailscale add-on implementations, and remembered somebody, somewhere asked this, so just wrote a quick comment. :)

george-viaud:
@lmagyar Thank you - that works flawlessly for my need.  I was able to use it in conjunction with nginx-proxy-manager plugin and can now address my server via a proper domain name even though that machine is behind a Dynamic IP that can not be rDNS'd due to ISP limitations.  Excellent.

Issue #67, opened 2022-01-02
Outbound access


pittbull:
# Problem/Motivation

Access to Glances running on a node in my Tailscale network

## Expected behavior

After installing the addon I expected HA to connect to this host successfully

## Actual behavior

Unable to access the host. HA shows in the TS admin and other Tailscale hosts can ping and access HA - I am not able though to access anything from HA.

I am not sure if this is supposed to work but I am unable to conclude based on the information I find.

robroll:
I am looking for a solution to this too.

My HA is on 192.168.5.x in one location. In another location I have a camera on 192.168.1.x connected to my Tailscale via an openwrt router. My laptop is able to access both. The HA (192.168.5.x) is unable to see the camera (I have created a ping sesnsor in the configuration file to test).

apmillen:
I am also looking for something like this. I think tailscale would have to be installed on the OS rather than in a docker container so I'm not sure this will ever work.

pittbull:
Any feedback on this?

JordiMD92:
Having the same issue.
I have 3 Home assistants with Tailscale client on each. I would like to be able to access from the main one to the secondary one with remote home assistant to have the information centralized.

Currently using wireguard but i want get rid of it.

frenck:
Currently, this add-on is accessing just Home Assistant itself. Due to its containerized nature and the way it is set up, it can currently not be used for routing Home Assistant itself out.

I'll make a task to update to documentation to make that more clear.

PR #170, opened 2023-04-03
Enable Tailscale's builtin outbound SOCKS5 and HTTP proxy

lmagyar:
# Proposed Changes

Tailscale runs in userspace-networking mode, accessing nodes in the Tailscale VPN from HA or other containers is not possible.

Though Tailscale provides SOCKS5 and HTTP proxy exactly for this situation (https://tailscale.com/kb/1112/userspace-networking/)

This PR optionally enables this functionality for those who want to access nodes on their Tailscale VPN.

## Related Issues

#169 #129 #67

Fixes: #169

### DIFF
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -54,6 +54,11 @@ subnets on all supported interfaces to Tailscale.
 Consider disabling key expiry to avoid losing connection to your Home Assistant
 device. See [Key expiry][tailscale_info_key_expiry] for more information.
 
+To allow Home Assistant and other add-ons to access your Tailscale network the
+add-on also provides a SOCKS5/HTTP proxy. The proxy protocol is unauthenticated.
+See [Userspace networking][tailscale_info_userspace_networking] for more
+information.
+
 ```yaml
 tags:
   - tag:example
@@ -67,16 +72,6 @@ This option allows you to specify specific ACL tags for this Tailscale
 instance. They need to start with `tag:`.
 
 More information: <https://tailscale.com/kb/1068/acl-tags/>
-
-### Option: `proxy_port`
-
-Optionally select a port on localhost (`127.0.0.1`) to listen on for connections
-from SOCKS5 and HTTP proxy-speaking applications. Enabling this feature allows
-Home Assistant and other add-ons to access your Tailscale network.
-
-The proxy protocol is unauthenticated.
-
-More information: [Userspace networking][tailscale_info_userspace_networking]
 
 ### Option: `log_level`
 
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -68,6 +68,14 @@ instance. They need to start with `tag:`.
 
 More information: <https://tailscale.com/kb/1068/acl-tags/>
 
+### Option: `proxy_port`
+
+Optionally select a port on localhost (`127.0.0.1`) to listen on for connections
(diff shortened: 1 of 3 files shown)


tailscale/config.yaml
```diff
@@ -27,4 +27,5 @@ map:
   - share:rw
 schema:
   tags: ["match(^tag:[a-zA-Z0-9]-?[a-zA-Z0-9]+$)?"]
+  proxy_port: match(^(?:[1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$)?
```
frenck, line 30: We should not add-on configuration for this; instead, assign a port for it. A user can change such ports from the UI.

frenck (changes requested):

lmagyar:
Modified it as required and rebased. And tested on real devices also.

tailscale/DOCS.md
```diff
@@ -54,6 +54,11 @@ subnets on all supported interfaces to Tailscale.
 Consider disabling key expiry to avoid losing connection to your Home Assistant
 device. See [Key expiry][tailscale_info_key_expiry] for more information.
 
+To allow Home Assistant and other add-ons to access your Tailscale network the
+add-on also provides a SOCKS5/HTTP proxy. The proxy protocol is unauthenticated.
+See [Userspace networking][tailscale_info_userspace_networking] for more
+information.
```
frenck, line 60: This is actually not usable as such. One won't be able to leverage the socks proxy for the described services.

tailscale/config.yaml
```diff
@@ -21,6 +21,10 @@ hassio_api: true
 host_network: true
 map:
   - share:rw
+ports:
+  1080/tcp: 1080
+ports_description:
+  1080/tcp: SOCKS5 and HTTP proxy port for outbound connections into your Tailscale network (make empty to disable)
```
frenck, line 27: Port descriptions should be in the translation files (this works, but is old-school).

tailscale/rootfs/etc/s6-overlay/s6-rc.d/tailscaled/run
```diff
@@ -14,6 +15,13 @@ options+=(--statedir=/data/state)
 options+=(--no-logs-no-support)
 options+=(--tun=userspace-networking)
 
+# If outbound proxy port is specified, enable outbound proxy
+outbound_proxy_port=$(bashio::addon.port 1080)
+if bashio::var.has_value "${outbound_proxy_port}"; then
+  options+=(--socks5-server=0.0.0.0:${outbound_proxy_port})
```
frenck, line 21: This add-on runs on the host network, and setting it to 0.0.0.0 will try to bind it on every interface it has available. I would recommend not doing that, as a conflict with the many many (virtual) interfaces, is hard to avoid.

frenck (changes requested):

tailscale/rootfs/etc/s6-overlay/s6-rc.d/tailscaled/run
lmagyar: Yes, and when you can merge the "drop userspace networking" branch, it will be useless.

Should I close this PR?

lmagyar:
As userspace networking is dropped, and there is a `tailscale0` interface now, I close this PR, because it is completely unnecessary.

frenck:
👍

PR #172, opened 2023-04-04
Extend documentation with disabling key expiry

lmagyar:
# Proposed Changes

Based on Tailscale's recommendation it is better to disable key expiry on servers/IoT devices. Without regular reauthentication users can loose connection to their devices permanently.

### DIFF
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -51,6 +51,9 @@ Tailscale account. Additionally, if the Supervisor managed your network (
 which is the default), the add-on will also advertise routes to your
 subnets on all supported interfaces to Tailscale.
 
+Consider disabling key expiry to avoid loosing connection to your Home Assistant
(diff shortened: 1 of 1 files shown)


tailscale/DOCS.md
```diff
@@ -51,6 +51,9 @@ Tailscale account. Additionally, if the Supervisor managed your network (
 which is the default), the add-on will also advertise routes to your
 subnets on all supported interfaces to Tailscale.
 
+Consider disabling key expiry to avoid loosing connection to your Home Assistant
+device. See [Key expiry][tailscale_info_key_expiry] for more information.
```
frenck, line 55:
```suggestion
Consider disabling key expiry to avoid losing connection to your Home Assistant
device. See [Key expiry][tailscale_info_key_expiry] for more information.
```

frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #184, opened 2023-04-15
Fix login-server option

reey:
# Proposed Changes

It should have been `login-server` instead of `login_server`.

Tested this and works fine now.


frenck (approved):
This is going well 😓 

Anyways, thanks for helping out! Appreciated 🙏 

../Frenck

PR #180, opened 2023-04-14
Add custom control server support

frenck:
# Proposed Changes

Adds support for setting a custom control server (e.g., a self-hosted Headscale instance).

supersedes and closes #175

PR #188, opened 2023-04-20
Add taildrop to example configuration

lmagyar:
# Proposed Changes

PR #185 didn't add taildrop to the example config in the DOCS.

Also PR #180 moved the tags config to the end in the description, but not in the example config.


frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #196, opened 2023-04-29
Remove duplicate status checks from dependent S6 services

lmagyar:
# Proposed Changes

Minor code quality improvement.

The `proxy` and `taildrop` services are waiting for `post-tailscaled`, so move the status checks from the beginning of `proxy` and `taildrop` to the end of `post-tailscaled`.

PR #181, opened 2023-04-14
Drop userspace networking

frenck:
# Proposed Changes

Drops userspace networking.

Needs:
- https://github.com/home-assistant/supervisor/pull/4242
- https://github.com/frenck/action-addon-linter/pull/137

PR #199, opened 2023-05-10
Make userspace networking configurable

lmagyar:
Note: this is based on / continuation of #196, but I will rebase this PR as required.

# Proposed Changes

The currently unreleased #181 dropped userspace networking automatically.

I've merged it into my fork, and 1 user complained, that his host lost all network access (even local 192.168.x.x did not worked). ***UPDATE:*** _The reason must be that the same subnet is advertised at multiple locations and it caused collision. The PR #201 provides a possible solution for collision, but I think making userspace-networking configurable still makes sense._

This PR makes it configurable, by default the add-on still uses userspace networking, but if somebody adds `userspace_networking: false` to the configuration, it will drop it and create the `tailscale0` interface with all of the dangers of it. 

But at least it requires manual intervention and doesn't happen automatically during an add-on update.

### DIFF
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -196,11 +196,11 @@ accessible within your tailnet.
 When not set, this option is enabled by default.
 
 If you need to access other clients on your tailnet from your Home Assistant
-instance, disable userspace networking mode, that will create a `tailscale0`
+instance, disable userspace networking mode, which will create a `tailscale0`
 network interface on your host.
 
 If you want to access other clients on your tailnet even from your local subnet,
-execute Step 2 and 3 as described on [Site-to-site
+execute steps 2 and 3 as described on [Site-to-site
 networking][tailscale_info_site_to_site].
 
 ### Option: `proxy`
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -70,6 +70,7 @@ tags:
   - tag:example
   - tag:homeassistant
 taildrop: true
+userspace_networking: true
 ```
 
 ### Option: `accept_dns`
@@ -185,6 +186,22 @@ This option lets you specify you to specify a custom control server instead of
 the default (`https://controlplane.tailscale.com`). This is useful if you
 are running your own Tailscale control server, for example, a self-hosted
 [Headscale] instance.
+
+### Option: `userspace_networking`
+
+The add-on uses [userspace networking mode][tailscale_info_userspace_networking]
+to make your Home Assistant instance (and optionally the local subnets)
+accessible within your tailnet.
+
+When not set, this option is enabled by default.
+
+If you need to access other clients on your tailnet from your Home Assistant
+instance, disable userspace networking mode, that will create a `tailscale0`
+network interface on your host.
+
+If you want to access other clients on your tailnet even from your local subnet,
+execute Step 2 and 3 as described on [Site-to-site
+networking][tailscale_info_site_to_site].
 
 ### Option: `proxy`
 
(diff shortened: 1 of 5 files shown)


frenck:
I have one instances that has the same problem (which is the main reason for it not being released yet).

However, I don't think this a solution nor a workaround. In some situations loosing access is problematic and would also prevent one from accessing this workaround.

../Frenck

Gyosa3:
> # Proposed Changes
> The currently unreleased #181 dropped userspace networking automatically.
> 
> I've merged it into my fork, and 1 user complained, that his host lost all network access (even local 192.168.x.x did not worked), the exact reason is under investigation currently.
> 
> This PR makes it configurable, by default the add-on still uses userspace networking, but if somebody adds `userspace_networking: false` to the configuration, it will drop it and create the `tailscale0` interface with all of the dangers of it.
> 
> But at least it requires manual intervention and doesn't happen automatically during an add-on update.
> 
> ## Related Issues

Hi I would like to understand what's behind the sentence "and create the `tailscale0` interface with all of the dangers of it."

According to Tailscale support team, enabling kernel mode with the `tailscale0`interface needs to be THE way to go to enable subnet routing. And this is what PR #181 tries to achieve and it fails for users who announce subnets in the addon.

So far I have seen no discussion, issue or pull requests that are addressing this general problem.

I'm trying to get to the root cause of it but it's very hard to reach out to anyone familiar with the combination of Linux kernet networking + docker networking + tailscale tunneling.

For reference the issues already open on this matter:
https://github.com/tailscale/tailscale/issues/8370
https://github.com/hassio-addons/addon-tailscale/issues/216
https://github.com/hassio-addons/addon-tailscale/issues/209

I am using this addon that provides additional configuration parameters:
https://github.com/tsujamin/hass-addons/tree/main/tailscale

And I'm trying to fork it out towards a working solution without userspace-networking so far to no avail.
I'd greatly appreciate your feedback!

lmagyar:
not stale, also multiple PR-s are based on this

lmagyar:
not stale

frenck:
Needs #201 first (and a rebase).

frenck:
>  but I think making userspace-networking configurable still makes sense.

I'm not sure why? I think we should just drop it fully.

../Frenck

tailscale/DOCS.md
```diff
@@ -186,6 +187,22 @@ the default (`https://controlplane.tailscale.com`). This is useful if you
...
+### Option: `userspace_networking`
+
+The add-on uses [userspace networking mode][tailscale_info_userspace_networking]
+to make your Home Assistant instance (and optionally the local subnets)
+accessible within your tailnet.
+
+When not set, this option is enabled by default.
+
+If you need to access other clients on your tailnet from your Home Assistant
+instance, disable userspace networking mode, that will create a `tailscale0`
```
frenck, line 199:
```suggestion
instance, disable userspace networking mode, which will create a `tailscale0`
```

tailscale/DOCS.md
```diff
@@ -186,6 +187,22 @@ the default (`https://controlplane.tailscale.com`). This is useful if you
...
+network interface on your host.
+
+If you want to access other clients on your tailnet even from your local subnet,
+execute Step 2 and 3 as described on [Site-to-site
```
frenck, line 203:
```suggestion
execute steps 2 and 3 as described on [Site-to-site
```

tailscale/translations/en.yaml
```diff
@@ -54,3 +54,11 @@ configuration:
...
       that allows you to share files with other Tailscale nodes.
       When not set, this option is enabled by default.
+  userspace_networking:
+    name: Userspace networking mode
+    description: >-
+      This option allows you to enable userspace networking mode.
+      If you need to access other clients on your Tailnet from your Home
+      Assistant instance, disable userspace networking mode, that will create a
+      `tailscale0` network interface on your host.
+      When not set, this option is enabled by default.
```
frenck, line 64:
```suggestion
      This option allows you to enable userspace networking mode.
      If you need to access other clients on your Tailnet from your Home
      Assistant instance, disable userspace networking mode, which will create a
      `tailscale0` network interface on your host.
      When not set, this option is enabled by default.
```

frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

lmagyar:
> > but I think making userspace-networking configurable still makes sense.
> 
> ~I'm not sure why? I think we should just drop it fully.~
> 
> Let's just do it.
> 
> ../Frenck

Yes, I just afraid that turning off userspace networking can mess up configs. Most of the people will not use tailscale0/tunneling I think.

PR #201, opened 2023-05-11
Protect local subnets from being routed toward Tailscale subnets if they collide

lmagyar:
Note: this is based on / continuation of #199, but I will rebase this PR as required.

# Proposed Changes

TLDR:
- when tailscaled starts: `ip rule add to "${route}" priority 5000 table main` for each advertised local subnets
- when stops: `ip rule del to "${route}"` for each previously added subnets
- when nm-dispatcher notifies about networking config change: delete rules and recalculate again

Idea is stolen from: https://github.com/tailscale/tailscale/issues/6231#issuecomment-1418124609 (Workaround 3)

Reason: Tailscale adds it's own routes with higher priority than the "normal" route settings, so we add back the local subnets with even higher priority. Tailscale wants to forward the local domain in case it is a coffee shop WiFi that collides with our tailnet settings, but in a server's case, it causes issues.

When Tailscale provides a solution for this old issue (https://github.com/tailscale/tailscale/issues/1227), we can remove this workaround.

`ip rule list` when the add-on is active/started:
```
0:      from all lookup local
5000:   from all to 192.168.1.0/24 lookup main                 <- added by this PR
5210:   from all fwmark 0x80000/0xff0000 lookup main
5230:   from all fwmark 0x80000/0xff0000 lookup default
5250:   from all fwmark 0x80000/0xff0000 lookup unspec unreachable
5270:   from all lookup 52
32766:  from all lookup main
32767:  from all lookup default
```

Tested on real HA OS, RPI 3 config, it didn't lost its network access when there were active colliding subnet routing, ~though I didn't make a "negative" test, that without this PR my RPI gets "bricked" with the same config~. UPDATE: I made this "negative" test unintentionally, so this PR really protects the access to the device.

### DIFF
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -204,10 +204,10 @@ execute steps 2 and 3 as described on [Site-to-site
 networking][tailscale_info_site_to_site].
 
 In case your local subnets collide with subnet routes within your tailnet, your
-local network access has priority and these addresses won't be routed toward
-your tailnet. This will prevent your Home Assistant instance to lose network
-conection. This also means that using the same subnet on multiple nodes for load
-balancing and failover is not possible with the current add-on behavior.
+local network access has priority, and these addresses won't be routed toward
+your tailnet. This will prevent your Home Assistant instance from losing network
+connection. This also means that using the same subnet on multiple nodes for load
+balancing and failover is impossible with the current add-on behavior.
 
 ### Option: `proxy`
 
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -203,6 +203,12 @@ If you want to access other clients on your tailnet even from your local subnet,
 execute steps 2 and 3 as described on [Site-to-site
 networking][tailscale_info_site_to_site].
 
+In case your local subnets collide with subnet routes within your tailnet, your
+local network access has priority and these addresses won't be routed toward
+your tailnet. This will prevent your Home Assistant instance to lose network
+conection. This also means that using the same subnet on multiple nodes for load
+balancing and failover is not possible with the current add-on behavior.
+
 ### Option: `proxy`
 
 When not set, this option is enabled by default.
--- a/tailscale/config.yaml
+++ b/tailscale/config.yaml
@@ -19,6 +19,7 @@ arch:
 init: false
 hassio_api: true
 host_network: true
+host_dbus: true
 privileged:
   - NET_ADMIN
   - NET_RAW
--- a/tailscale/Dockerfile
+++ b/tailscale/Dockerfile
@@ -16,6 +16,7 @@ RUN \
         iptables=1.8.9-r2 \
         nginx=1.24.0-r6 \
         coreutils=9.3-r1 \
+        networkmanager-common=1.42.8-r0 \
     \
     && ln -sf /sbin/xtables-nft-multi /sbin/ip6tables \
     && ln -sf /sbin/xtables-nft-multi /sbin/iptables \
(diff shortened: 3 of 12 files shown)


lmagyar:
I've added a warning into the log, that there are colliding subnets when the add-on starts (this is independent of the permanently active, higher priority ip rule redirection). Tested on 3 real devices, seems to be solid.

lmagyar:
Fixed IPv6 handling.

lmagyar:
Added handling local network changes (eg. from the UI changing IP, mask, etc.). Moved the functionality to a new s6 service.

Eg. when NetworkManager reported the manual removal of IPv6 IP address, the ip rules (protecting the local subnets from collision/redirection) got recalculated.

```
    [21:44:32] INFO: Removing route 192.168.1.0/24 from ip rules
    [21:44:32] INFO: Removing route fd19:8efe:2a99::/64 from ip rules
    [21:44:36] INFO: Adding local subnets to ip rules with higher priority than Tailscale's routing,
    [21:44:36] INFO: to prevent routing local subnets if the same subnet is routed within your tailnet.
    [21:44:36] INFO: Adding route 192.168.1.0/24 to ip rules
```

ss89:
Would love to see this merged to try it out. Do we need to make @frenck aware of this? Or is there other blockers?

lmagyar:
I'm sure he is aware, but he has limited time. :(

You can try/test it by installing my fork until a solution is provided in the official add-on: https://github.com/lmagyar/homeassistant-addon-tailscale

tailscale/config.yaml
```diff
@@ -19,6 +19,7 @@ arch:
 init: false
 hassio_api: true
 host_network: true
+host_dbus: true
```
frenck, line 22: This is a very unwanted addition for add-ons. We should remove the dependency on the dbus.

tailscale/DOCS.md
```diff
@@ -203,6 +203,12 @@ If you want to access other clients on your tailnet even from your local subnet,
 execute steps 2 and 3 as described on [Site-to-site
 networking][tailscale_info_site_to_site].
 
+In case your local subnets collide with subnet routes within your tailnet, your
+local network access has priority and these addresses won't be routed toward
+your tailnet. This will prevent your Home Assistant instance to lose network
+conection. This also means that using the same subnet on multiple nodes for load
+balancing and failover is not possible with the current add-on behavior.
```
frenck, line 210:
```suggestion
In case your local subnets collide with subnet routes within your tailnet, your
local network access has priority, and these addresses won't be routed toward
your tailnet. This will prevent your Home Assistant instance from losing network
connection. This also means that using the same subnet on multiple nodes for load
balancing and failover is impossible with the current add-on behavior.
```

tailscale/config.yaml
frenck: I see... we need this.

frenck (commented):
Made a small tiny docs tweak. Otherwise: Beautiful!

frenck (approved):

lmagyar:
> Made a small tiny docs tweak. Otherwise: Beautiful!

Thank you. :)

Though it has 2 logical bugs/issues:
1. **Real bug:** It protects the **advertised** subnets, what was originally identical with the **local** subnets, but after the advertised subnets are user configurable, I think this stuff should always protect **local** subnets, independently of what users configured to be advertised. Because collisions will be between local and accepted routes not between advertised and accepted. It's only a few lines of code, I can make a PR.
2. **Maybe not a bug:** It protects subnet domains, may be should only protect HA's local IPs only. But I'm not sure whether is solves or creates more problems.

I think 1. needs a new PR, 2. can stay as is. What do you think?

PR #242, opened 2023-08-09
Update Installation section in documentation

lmagyar:
# Proposed Changes

Some services are starting after a successful login (eg. taildrop, proxy), and users should check the logs again after login, that these services are started successfully also.

Also added a step to check HA configuration for proxy function before starting the add-on. This is important, because without proper `trusted_proxies` configuration, the add-on won't work.


frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #252, opened 2023-09-01
make accepting subnet routes optional

willnorris:
This is needed in some scenarios, such as when the other node is advertising the subnet that home assistant is on.

### DIFF
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -56,6 +56,7 @@ device. See [Key expiry][tailscale_info_key_expiry] for more information.
 
 ```yaml
 accept_dns: true
+accept_routes: true
 advertise_exit_node: true
 log_level: info
 login_server: "https://controlplane.tailscale.com"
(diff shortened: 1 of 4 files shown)


lmagyar:
Just a note (I'm only a plain user of this add-on, and a maintainer of a fork of it): users still can misconfigure their setup, and that can "soft brick" their HA device, and then they are unable to disable "accepting subnet routes". (And first they have to understand what caused the "soft bricking" in the first place.)

So there is PR #201 that is a hacky workaround to prevent the "soft bricking", and that emits some information about the situation into the log. Though that prevents this add-on to be used for load balancing and failover.

To be honest, I don't know what is a good solution for this situation. I think:
- we need both PR
- maybe PR #201 should protect only the HA device's local IP not the whole local network

lmagyar:
not stale

frenck (approved):
Thanks, @willnorris 👍 

../Frenck

PR #253, opened 2023-09-07
Make advertise routes configurable

lmagyar:
Note: this is based on / continuation of #201, but I will rebase this PR as required.

# Proposed Changes

Make it possible to advertise not only the default routes on the managed interfaces, but user defined routes also, to support advanced networking situations (like site-to-site networking with multiple subnets, traverse multiple networks, etc.).

## Related Issues


frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #254, opened 2023-09-08
Test Home Assistant's HTTP reverse proxy configuration on add-on startup

lmagyar:
# Proposed Changes

This will prevent "400: Bad Request" errors and redirects users to the Documentation.


tailscale/rootfs/etc/s6-overlay/s6-rc.d/proxy/run
```diff
@@ -31,6 +31,13 @@ if bashio::var.true "$(bashio::core.ssl)"; then
   bashio::exit.nok
 fi
 
+# Test Home Assistant's HTTP reverse proxy configuration
+if (( 200 != $(curl -s -o /dev/null -w "%{http_code}" "http://127.0.0.1:$(bashio::core.port)" -H "X-Forwarded-For: 127.0.0.1") )); then
```
frenck, line 35: Home Assistant doesn't have to be running?

frenck (changes requested):

tailscale/rootfs/etc/s6-overlay/s6-rc.d/proxy/run
```diff
@@ -31,6 +31,13 @@ if bashio::var.true "$(bashio::core.ssl)"; then
...
+  bashio::log.error "Unable to connect to Home Assistant on http://127.0.0.1:$(bashio::core.port)"
```
frenck, line 36: Let's remove the technical detail here, it can be confusing as lots of users won't care/know what `127.0.0.1` is doing here.

```suggestion
  bashio::log.error "Unable to connect to Home Assistant."
```

tailscale/rootfs/etc/s6-overlay/s6-rc.d/proxy/run
lmagyar: It tests that the `X-Forwarded-For: 127.0.0.1` is added to the HA configuration. People usually forget it and then see the "400: Bad Request" error and complain. so I've made a test and a log entry.

**It has a "dangerous" side effect:** if by default the proxy is enabled, but the user has never tried it and never noticed the misconfiguration, the add-on started previously, but with this change, it won't start, because the default enabled proxy is not working, so it will brake some systems, this happened once with one of "my" forked repo's users.

And it also means, the after a fresh add-on install, the add-on won't start, users have to reconfigure HA or disable the TS proxy. It's a bit strange, but follows your concept, to fail if there is something misconfigured.

frenck: I get what it does and why you did it, but there is no guarantee Home Assistant is even running when it reaches this point.

lmagyar: Oh, I see. `startup: services` will start before Home Assistant, though nobody run into this problem yet, the add-on starts up much slower then HA core.

Maybe emit the error, or only a warning, but do not `exit.nok`?

frenck: A little bit of a d-tour, maybe, but I can expose the core state so we can check if Core is running or not. This might help with the decision to run this check.

I can probably make this happen friday/the weekend.

lmagyar: That sounds great! If HA runs, current behavior (error), if doesn't, only a warning about it.

lmagyar:
Rebased, to prevent merge conflicts (proxy service code will change a lot).

tailscale/rootfs/etc/s6-overlay/s6-rc.d/proxy/run
frenck: Updated the base image, which contains a new method `bashio::info.state`, if the output of that equals to `running` you'd know Core is up!

frenck:
> Note: the unrelated changes are already in PR https://github.com/hassio-addons/addon-tailscale/pull/277, this PR needs to be rebased after that is merged.

Please don't stack your PRs on top of each other. Make them clean and standalone. Can you rebase this one and pick only the related commits? Thanks 👍 

../Frenck

tailscale/rootfs/etc/s6-overlay/s6-rc.d/proxy/run
lmagyar: Isn't that the Supervisor's state? It is constant "running", API docs says: "The core state of the Supervisor."

That field returns `self.sys_core.state`

We would need sg. like `self.sys_homeassistant.core.is_running`, but I can't find it on the API.

lmagyar:
And rebased also.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/proxy/run
frenck: > Isn't that the Supervisor's state?

You are so right :(

"core" in the Supervisor API threw me off. Will check with the team.

lmagyar: Made a workaround: first curl HA **without** `-H "X-Forwarded-For: ...` to test whether it is running (accessible) or not, if running, test with `-H "X-Forwarded-For: ...`

Tested, all 3 cases working:
- core is not runnng
- core is running without proper config
- core is running with proper config

Rebased, ready to review.

frenck (approved):
That seems like reasonable way to handle this 👍 

Thanks, @lmagyar 👍 

../Frenck

PR #260, opened 2023-09-25
Use new .Self.CapMap in status JSON for HTTPS support check

lmagyar:
# Proposed Changes

Status JSON has changed. Requires Tailscale v1.50.0 or later.

## Related Issue


frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #273, opened 2023-10-17
Proxy and Funnel are disabled by default

### LINKED ISSUE #272: Please turn funnel off by default and consider turning proxy off by default
While the funnel setting has no effect with funnel turned off on a tailnet level, I feel that an option which has the potential to expose a home-assistant installation to the entire internet should be opt-in. I found it rather an unpleasant surprise seeing that this was enabled by default as I saw this addon as a way to get **secure** remote access.
While proxy mode is a nice addition which will allow me to remove the separate nginx reverse proxy addon, having proxy enabled on this addon by default broke my existing remote access setup as it grabbed port 443 before I added the correct trusted proxies line. It was easy enough to fix having local access, but not the smoothest experience. 

Thanks for building this addon.

lmagyar:
## Proposed Changes

Nuke `tailscale serve` and `tailscale funnel` configs on startup (better nuke I think, who knows what change will come later).

This can be done only after `tailscale up` is ready. Then we have to stop the add-on if deleting the configs wasn't successful (small chance).

## Related Issues

fixes #272

### DIFF
--- a/tailscale/rootfs/etc/s6-overlay/s6-rc.d/funnel/run
+++ b/tailscale/rootfs/etc/s6-overlay/s6-rc.d/funnel/run
@@ -11,7 +11,7 @@ declare domain
 if ! /opt/tailscale status --self=true --peers=false --json \
   | jq -rce '.Self.CapMap | has("https")' > /dev/null;
 then
-  bashio::log.error "Tailscale's HTTPS support is disabled. Therefore, can't start the add-on's Tailscale Funnel functionality."
+  bashio::log.error "Tailscale's HTTPS support is disabled"
   bashio::exit.nok
 fi
 
@@ -21,7 +21,7 @@ domain=$(/opt/tailscale status --self=true --peers=false --json | jq -rc ".CertD
 if ! /opt/tailscale status --self=true --peers=false --json \
   | jq -rce '.Self.CapMap | has("funnel")' > /dev/null;
 then
-  bashio::log.error "Tailscale's Funnel support is disabled. Therefore, can't start the add-on's Tailscale Funnel functionality."
+  bashio::log.error "Tailscale's Funnel support is disabled"
   bashio::exit.nok
 fi
 
--- a/tailscale/rootfs/etc/s6-overlay/s6-rc.d/funnel/run
+++ b/tailscale/rootfs/etc/s6-overlay/s6-rc.d/funnel/run
@@ -11,8 +11,8 @@ declare domain
 if ! /opt/tailscale status --self=true --peers=false --json \
   | jq -rce '.Self.CapMap | has("https")' > /dev/null;
 then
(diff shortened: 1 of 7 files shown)


frenck (changes requested):
To fix the referenced issue, we also need to change the defaults.

lmagyar:
> To fix the referenced issue, we also need to change the defaults.

And rewrite the DOCS, translate.yaml, a lot of mentioning of the proxy functionality... I will do that a few hours later, but now I have to stand up from my keyboard...

lmagyar:
I've made both proxy and funnel by default disabled. Basic manual tests are done on my real rPI-s, enabling/disabling them, everything seems to be OK.

Positive side effect: after install the add-on can be started immediately, no HA reverse proxy or add-on config required.

lmagyar:
And from now on when proxy and funnel service can't start, this is not a notice, this is an error.

frenck:
[1e63fdb](https://github.com/hassio-addons/addon-tailscale/pull/273/commits/1e63fdb1df3856124a6e9360c1a606c32faeaa3d)

This is not related to this PR. Please move it.

frenck (changes requested):
Wrong button. See above comment.

lmagyar:
OK, but it makes really sense. Will make a PR based on this PR.

lmagyar:
Finally moved all non port related code into this PR, because now it's much clearer which service does what.
Now both are ready to review.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/funnel/run
```diff
@@ -5,31 +5,51 @@
...
+# Disable previously configured tailscale funnel functionality with deleting the tailscale funnel configuration
+# There is no tailscale funnel reset command currently, we loop through previous config
+for already_configured_funnel_port in $( \
+  /opt/tailscale funnel status --json \
+  | jq -rc '.AllowFunnel // [] | keys[]' \
+  | sed -nr 's/^[^:]*:(\d*)$/\1/p');
+do
+  if ! /opt/tailscale funnel "${already_configured_funnel_port}" off; then
+    bashio::log.error "Unable to reset Tailscale Funnel settings"
+    bashio::exit.nok
```
frenck, line 21: Wouldn't raising an `nok` on the service run cause the service to restart?

lmagyar: Yes, that was my goal, like in case of the proxy: if we can't reset the config, do not let `tailscale up` to run, because that will expose HA to funnel even when the user do not want it anymore. Sadly we can test the current proxy/funnel config only after `tailscale up` was successful.

frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #275, opened 2023-10-17
Fix local subnet protection

lmagyar:
## Proposed Changes

Fixes a logical bug in #201: That PR protected the ***advertised*** subnets, what was originally identical with the ***local*** subnets, but after the advertised subnets are user configurable, I think the subnet protection logic should always protect ***local*** subnets, independently of what users configured to be advertised. Because collisions will be between local and accepted routes not between advertised and accepted.

**Note:** There is another possible improvement, but that is not addressed by this PR: We protect subnet domains, maybe should only protect HA's local IPs only. But I'm not sure whether it solves or creates more problems. Until there is some concrete need for it, I prefer not to modify the current logic.

### DIFF
--- a/tailscale/rootfs/etc/s6-overlay/s6-rc.d/mss-clamping/run
+++ b/tailscale/rootfs/etc/s6-overlay/s6-rc.d/mss-clamping/run
@@ -9,7 +9,7 @@ declare -a routes=()
 declare -a interfaces=()
 declare route family interface
 
-readarray -t routes < <(subnet-routes)
+readarray -t routes < <(subnet-routes advertised)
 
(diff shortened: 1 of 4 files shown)


frenck (changes requested):
There are unrelated commits in this PR, please remove them.

lmagyar:
Rebased, to prevent merge conflicts.

frenck:
> Rebased, to prevent merge conflicts.

Can you please remove the unrelated commits? As per request above. Rebase only related changes please 👍

lmagyar:
And you are right again... (no merge conflict with other proxy modification in either way).
Marking ready to review.

frenck (approved):
Nice! Thanks, @lmagyar 👍 

../Frenck

PR #277, opened 2023-10-19
Make Proxy and Funnel port configurable

### LINKED ISSUE #276: Tailscale Add-on's embedded Nginx proxy intercepts all in-bound HTTPS traffic from tailnet
# Problem/Motivation

> Can no longer access Nginx Proxy Manager on same host as Tailscale

## Expected behavior

> Setting `funnel: false` and `proxy: false` in the YAML for the Tailscale Add-On should disable its internal Nginx proxy server and instead forward HTTP requests to the Nginx Proxy Manager Add-On on the host.

## Actual behavior

> Setting `funnel:false` and `proxy:false` in the UI and the YAML configuration for the add-on does nothing. It always resets itself to `true` on every restart of the Tailscale Add-On.

## Steps to reproduce

> 1. Install Nginx Proxy Manager add-on and set up reverse proxies for various services on the network. Contact these services remotely over the tailnet and they're fine.
> 2. Install Tailscale add-on and connect it to a tailnet with funnel enabled in the admin panel. Now you cannot access them anymore as the Nginx proxy inside the Tailscale add-on will catch them and say "I don't know what host that is" and fail.

## Proposed changes

> Make it so we can actually turn off funnel and/or proxy, completely disabling the Nginx proxy inside this add-on and resuming forwarding to the HA host device where Nginx Proxy should catch it on port 80/443.

OR

> Make it so I can instead set up an entire list of `tailscale serve` endpoints in the configuration for this add-on so I can use it instead of the Nginx Proxy Manager one.

lmagyar:
## Proposed Changes

Also made the port configurable. Proxy and Funnel are using the same port, so making 2 port configs and dropping the current `proxy: bool?` and `funnel: bool?` configs can't be done. So 443 by default is configured as port, but this won't start proxy or funnel automatically.

## Related Issues

fixes #276

### DIFF
--- a/tailscale/config.yaml
+++ b/tailscale/config.yaml
@@ -35,6 +35,7 @@ schema:
   log_level: list(trace|debug|info|notice|warning|error|fatal)?
   login_server: url?
   proxy: bool?
+  proxy_and_funnel_port: list(443|8443|10000)?
   snat_subnet_routes: bool?
   tags:
     - "match(^tag:[a-zA-Z0-9]-?[a-zA-Z0-9]+$)?"
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -67,6 +67,7 @@ funnel: false
 log_level: info
 login_server: "https://controlplane.tailscale.com"
 proxy: false
+proxy_and_funnel_port: 443
 snat_subnet_routes: true
 tags:
   - tag:example
@@ -158,9 +159,9 @@ More information: [Tailscale Funnel][tailscale_info_funnel]
 **Note**: _After initial setup, it can take up to 10 minutes for the domain to
 be publicly available._
 
-**Note:** _You should not use any port number in the URL that you used
+**Note:** _You should not use the port number in the URL that you used
 previously to access Home Assistant. Tailscale Funnel works on the default HTTPS
-port 443._
+port 443 (or the port configured in option `proxy_and_funnel_port`)._
 
 **Note:** _If you encounter strange browser behaviour or strange error messages,
 try to clear all site related cookies, clear all browser cache, restart browser._
@@ -234,9 +235,19 @@ More information: [Enabling HTTPS][tailscale_info_https]
 
 1. Restart the add-on.
 
-**Note:** _You should not use any port number in the URL that you used
+**Note:** _You should not use the port number in the URL that you used
 previously to access Home Assistant. Tailscale Proxy works on the default HTTPS
-port 443._
+port 443 (or the port configured in option `proxy_and_funnel_port`)._
+
+### Option: `proxy_and_funnel_port`
+
+This option allows you to configure the port the Tailscale Proxy and Funnel
+features are accessible on the tailnet (in case of Tailscale Proxy is enabled)
+and optionally on the internet (in case of Tailscale Funnel is also enabled).
+
+Only port number 443, 8443 and 10000 is allowed by Tailscale.
+
+When not set, port number 443 is used by default.
 
 ### Option: `snat_subnet_routes`
 
--- a/tailscale/rootfs/etc/s6-overlay/s6-rc.d/serve/run
+++ b/tailscale/rootfs/etc/s6-overlay/s6-rc.d/serve/run
@@ -63,4 +63,4 @@ else
 fi
 
 # Set up serve
-exec /opt/tailscale ${tailscale_command} --https=443 --set-path=/ "http://127.0.0.1:$(bashio::core.port)"
+exec /opt/tailscale ${tailscale_command} --https=$(bashio::config "proxy_and_funnel_port" "443") --set-path=/ "http://127.0.0.1:$(bashio::core.port)"
--- a/tailscale/translations/en.yaml
+++ b/tailscale/translations/en.yaml
@@ -50,6 +50,12 @@ configuration:
       This option allows you to enable Tailscale's Proxy feature to present your
       Home Assistant instance on your tailnet with a valid certificate.
       When not set, this option is disabled by default.
+  proxy_and_funnel_port:
+    name: Tailscale Proxy and Funnel port
+    description: >-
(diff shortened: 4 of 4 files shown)


tailscale/config.yaml
```diff
@@ -41,3 +41,7 @@ schema:
   tags: ["match(^tag:[a-zA-Z0-9]-?[a-zA-Z0-9]+$)?"]
   taildrop: bool?
   userspace_networking: bool?
+ports:
+  443/tcp: 443
+ports_description:
+  443/tcp: Tailscale Proxy and Funnel port
```
sinclairpaul, line 47: The addon is on the host network, so does it make sense to use the port mapping feature?, as it looks like this is then referenced in the scripts, would it make more sense to use a config option (where you could also define the list of valid options) ?

lmagyar: It's true, but it was the suggested way in similar situation (https://github.com/hassio-addons/addon-tailscale/pull/170#discussion_r1157057299) to configure ports even when we are on the host network (most users won't know the diference).

lmagyar:
Finally moved all non port related code into the #273 PR, because now it's much clearer which service and PR does what.
Now both are ready to review.

tailscale/config.yaml
lmagyar: And eg. the SSH add-on works this way also, I've stolen the idea from there. https://github.com/hassio-addons/addon-ssh/blob/main/ssh/rootfs/etc/s6-overlay/s6-rc.d/init-ssh/run

sinclairpaul: No worries, I get it, just as the requirement is ideally to have a selected port definition and in my mind ideally a default that isn't 443, it would make more sense.

However I will of course defer to Frenck :)

lmagyar:
Rebased it, and ready to review.

tailscale/config.yaml
frenck: I don't feel like this is used similar to what SSH does. While this does provide an interface, this is used the other way around. 

Not sure on this tbh 🤔

sinclairpaul: Just from a user perspective if we need to restrict the option set, we can really only do that from a list anyway, and as everything is now in a nice GUI, does it make sense to use the Port Mapping inputs?

lmagyar: I think in case of reverse proxy functionality, it works similar to SSH or NGINX or MQTT, this is a service published by the add-on.

Funnel is a tricky concept, yes. But Funnel can serve only the same ports (on the same URLs) that the reverse proxy publishes, so it is "only" publishes this service not only on the tailnet, but on the internet. So we can say, this is the same service, with wider network accessibility (like opening a port on the router won't change our concept of NGINX proxy functionality on port X).

frenck: I don't think that works, as one basically maps X:Y. In case of the UI offered by Home Assistant, the user can't configure X, only Y.

lmagyar: > I don't think that works, as one basically maps X:Y. In the case of the UI offered by Home Assistant, the user can't configure X, only Y. (And even that is in a limited way).

The current hard wired 443 value can collide with already existing services.

sinclairpaul: Apologies, I'm not that familiar with Tailscale, but I think I get it now.  The proxy can be configured independently, and provides a reverse proxy on the tailnet which can run on any configured port.  I'm assuming that the port isn't availiable outside of the tailnet? (i.e. on the host)

Funnel is then a configured item that exposes via Tailscale's services (hence the port requirement), which then advertises via the Internet.

If they are independent, why are they referencing the same config item?

I would use something along the lines of:

Proxy - Port definition (although I think it's not really exposing a port)

Funnel - List of Nope, 443, 8443, 10000

lmagyar: Yes, proxy can run on any port inside the tailnet.

> If they are independent, why are they referencing the same config item?

No, they are not independent.

Funnel can "choose" 3 ports (443, 8443, 10000) to publish on the internet. But those 3 ports have to be already existing Proxy ports, so "funneling" is only a "yes/no" decision on the user's side, he can't say, to run Proxy on port 9876 and publish it on the internet as 8443, the 2 ports has to be the same.

lmagyar:
Not stale, decision is to be made:
- move this option from port config to plain config like `proxy_and_funnel_port: port?` (because this is not a real host port, it is available only in the tailnet)
- or forget it in favor of PR #293 (which I think is a fine PR, but I think it is a big step for an average user to use command line to configure a basic thing like a port, so I prefer to move this option to a plain option and leave advanced command line for more experienced users)

frenck:
I actually think we need to move this to an add-on option, which also allows us to guard that the port number is in one of the accepted ones.

I feel like this is a bit too much abusing a port feature that was not meant for this usecase (and allows for any port, even if it doesn't make sense).

../Frenck

frenck:
> which allows using the cli for serve config

We should never expose that.

lmagyar:
I've used `proxy_and_funnel_port: list(443|8443|10000)?` option to configure it.

Issues:
- This limits the port number to this 3, even when only Proxy (tailscale serve) is used, though this is a Funnel restriction. But I think this will cause less complaints than enabling any port for Proxy that runs on to error when the user later enables Funnel.
- The generated YAML configuration looks a bit ugly for an integer port number: `proxy_and_funnel_port: "8443"`
- If the user edit the YAML and enters `proxy_and_funnel_port: "8443"` the UI will not give an error, but the non-YAML view can't select the right "8443" from the list. The add-on works certainly. This seems to be a UI bug around int/str values.

Otherwise ready to review.

frenck:
Hmm right, the list is considered a list of string by the SU.

frenck (approved):
Let's see how it behaves. Thanks, @lmagyar 👍 

../Frenck

lmagyar:
Wait, wait!

We can use a regex!

lmagyar:
Give me a few minutes to test it.

PR #280, opened 2023-10-21
Optionally copy Tailscale Proxy's certificate files to /ssl folder

lmagyar:
## Proposed Changes

Optionally exports the `tailscale serve` ie. Tailscale Proxy's certificates to the `/ssl` folder, like the DuckDNS add-on.

***Note:*** There is a HA UI/Supervisor bug (https://github.com/home-assistant/supervisor/issues/2640) that makes the logical configuraton options impossible to use, that's why currently the options are a bit ugly. This can be reverted later with one click. So, instead of this:

```
lets_encrypt:
  certfile: str?
  keyfile: str?
```

This is used:

```
lets_encrypt_certfile: str?
lets_encrypt_keyfile: str?
```

## Related Issues

fixes feature request reported in my experimental forked repo https://github.com/lmagyar/homeassistant-addon-tailscale/issues/68


frenck:
Let's not go this path. If anything, we are moving towards removing SSL natively from HA (at least, that is an end goal).

../Frenck

lmagyar:
The Tailscale Proxy certificate export wasn't for HA core itself, it was for other add-ons.

The user who requested it said: "I use it for the Adguard Home add-on's encryption feature so I can use my Android phone's "private DNS" feature."

I don't know/use AdGuard at all, but isn't it a valid use case?

frenck:
> The Tailscale Proxy certificate export wasn't for HA core itself, it was for other add-ons.

Those will have a similar faith in the long run. I've already start removing that feature from add-ons.

PR #274, opened 2023-10-17
Documentation improvements

lmagyar:
## Proposed Changes

Typos, consistent formatting, consistent wording, alphabetical order after the mega-merge in v0.12.0.

~One functional improvement: Protect local subnets only if non-userspace networking is used ***and accepting routes are enabled***.~

PR #283, opened 2023-10-23
Protect local subnets only if accepting routes are enabled

lmagyar:
# Proposed Changes

Protect local subnets only if non-userspace networking is used ***and accepting routes are enabled***.

Small code improvement moved into separate PR from #274.


frenck (approved):

frenck:
Thanks, @lmagyar 👍 

../Frenck

PR #288, opened 2023-10-31
Use modified tailscale cli arguments for proxy

### LINKED ISSUE #287: Tailscale (breaking) changes for serve and funnel cli in v1.52.0
UPDATE: With new PR #306 using tailscale serve's new arguments we can run proxy and funnel functionality in a longrun service and we don't need to reset serve config on each start (it also eliminates the need for an advanced_config option to disable this reset).

lmagyar:
# Proposed Changes

SSIA

## Related Issues

partially fixes #287


lmagyar:
Closing in favor of a better solution.

PR #293, opened 2023-11-11
Advanced Tailscale Proxy and Funnel configuration

lmagyar:
# Proposed Changes

Enable users to make custom tailscale serve and funnel configs. Only a 1 bit config, and nearly nothing to do in the add-on's code.

When enabled, all proxy and funnel manipulation is disabled in the add-on. Tailscale serve/funnel will remember the settings on their own:
- previously created by the add-on, or
- by the user executing tailscale serve/funnel commands, or
- by even the not-yet-existing tailscale UI.

## Related Issues

Closes PR #277 #278 Issue #276

### DIFF
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -118,7 +118,7 @@ working properly.
 Recommended steps to configure:
 
 1. Login to this add-on's container with
-   `docker exec -it addon_09716aab_tailscale /bin/bash`
+   `docker exec -it addon_a0d7b954_tailscale /bin/bash`
 
 1. Fine tune your tailscale settings with manual `/opt/tailscale serve --bg ...` and
    `/opt/tailscale funnel --bg ...` commands.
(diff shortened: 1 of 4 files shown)


tailscale/DOCS.md
```diff
@@ -96,6 +97,36 @@ More information: [Subnet routers][tailscale_info_subnets]
...
+
+This option is for advanced users who really know what they are doing. Though it
+is recommended even for them, to set up proxy and funnel at first, and only
+start advanced manual configuration when the basic proxy and funnel features are
+working properly.
+
+Recommended steps to configure:
+
+1. Login to this add-on's container with `docker exec -it
+   addon_09716aab_tailscale /bin/bash`
```
daalbert: A more robust method:
```docker exec -it `docker ps -q -f "name=tailscale"` /bin/bash```

lmagyar: Thank you, I've modified it.

lmagyar:
Closing in favor of a better solution.

PR #306, opened 2023-12-16
Use new tailscale cli arguments and merge proxy and funnel into longrun serve service

### LINKED ISSUE #287: Tailscale (breaking) changes for serve and funnel cli in v1.52.0

lmagyar:
# Proposed Changes

Using tailscale serve's new arguments we can run proxy and funnel functionality in a longrun service and we don't need to reset serve config on each start (it also eliminates the need for an advanced_config option to disable this reset).

It merges all the functionality from #288 #289 #290 #293 in a simpler way.

## Related Issues

closes #287


tailscale/rootfs/etc/s6-overlay/scripts/stage2_hook.sh
```diff
@@ -25,3 +25,8 @@ fi
 if bashio::config.false 'taildrop'; then
     rm /etc/s6-overlay/s6-rc.d/user/contents.d/taildrop
 fi
+
+# Disable serve service when proxy and/or funnel has not been explicitly enabled
+if ! bashio::config.true 'proxy'; then
```
elcajon, line 30: I guess you need to check for proxy OR funnel = true if the service handles both

QuinnyPig:
Looks like the fix to Home Assistant's tailscale impacting routing got deployed today. Can we unstick this?

lmagyar:
Also added some wait for HA during startup, because even when `tailscale serve/funnel` can start up without HA and can use it later, in case of userspace networking enabled tailscale will fill the log with IPv6 related messages in each ~10s forever (even after HA started up and the serve/funnel connection is working, seems to be a TS bug).

The wait for HA is only max. 90s, so the add-on will continue eventually and will try to start up the `tailscale serve/funnel` connection as previously (maybe this is the only connection to the device, better to try than not).

tailscale/DOCS.md
```diff
@@ -165,6 +165,15 @@ port 443._
...
 
+**Note**: _Advanced users who really know what they are doing can configure
+tailscale's proxy and funnel feature directly from the command line. Though it
+is recommended even for them, to set up proxy and funnel at first through the
+add-on options, and only start advanced manual configuration when the basic
+proxy and funnel features are working properly. To login to this add-on's
+container use ``docker exec -it `docker ps -q -f name=tailscale` /bin/bash``, to
+fine tune your tailscale settings use the `/opt/tailscale serve --bg ...` and
+`/opt/tailscale funnel --bg ...` commands._
+
```
frenck, line 176:
```suggestion
```

tailscale/DOCS.md
```diff
@@ -238,6 +247,15 @@ More information: [Enabling HTTPS][tailscale_info_https]
...
```
frenck, line 258:
```suggestion
```

frenck (approved):
Removed the two sections, to make it ready to merged. We can continue discussion on it (even after the merge) in case you feel like it 👍 

../Frenck

lmagyar:
This is absolutely fine, I've deleted the same paragraphs, but you pushed it faster. :)

PR #307, opened 2023-12-16
Cleanup ip6tables

lmagyar:
# Proposed Changes

Base image v15 also bumped alpine version.

Also remove ip6tables reference, because it is part of the plain iptables from v3.19.


frenck:
Merged with lastest `main`, adjusted the title, as the only thing left is the removal of ip6tables (after #302)

frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #308, opened 2023-12-16
Clamp the MSS for IPv6 also

lmagyar:
# Proposed Changes

Bugfix: Clamping was done for IPv4 only, this adds it for IPv6 traffic also.


frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #309, opened 2023-12-21
Increase wait time for Supervisor

lmagyar:
# Proposed Changes

Just a small improvement.

Currently it waits only 30s, but after a SU update the restart can take easily 10 seconds, so increase it from 30 (15 x 2s) to 90s (18 x 5s).

```
[14:45:36] INFO: Removing route 192.168.1.0/24 from ip rules
[14:45:37] INFO: Waiting for the supervisor to be ready...
[14:45:40] INFO: Waiting for the supervisor to be ready...
[14:45:42] INFO: Waiting for the supervisor to be ready...
[14:45:44] INFO: Supervisor is ready
[14:45:45] INFO: Adding local subnets to ip rules with higher priority than Tailscale's routing,
```


frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #329, opened 2024-02-09
Use web as a command, not as an option

lmagyar:
# Proposed Changes

Minor code quality improvement. No functional change.

Use `web` as `file`, `up`, `serve`, etc.


frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #330, opened 2024-02-09
Use readonly webui mode in v1.60

### LINKED ISSUE #314: Read-only Tailscale web UI
I see 3 related issues, but I don't know the solution. I think this should be better fixed with a workaround for HA in `tailscale web` code, that can eliminate the below mentioned http request and solve all issues. But I'm a total newbie in case of web development, so maybe somebody knows better.

## 1. How TS identifies the browser/user

As I see, some script running in the browser (when you press the "Viewing" button) tries to directly access `http://<device's TS IP>:5252/ok`, and when the browser is not part of the tailnet, certainly it fails. I see some workaround in the `tailscale web` source for synology and qnap to accept their user authentication.

So without a TS workaround for HA, I see 0 chance to access the non-read-only UI if the browser is not part of the tailnet but the user is an admin HA user.

## 2. Non-read-only UI opens up in a new tab/window

When I access HA's plain ***http*** UI (browser -> HA/Ingress (http) -> add-on's nginx -> tailscale web), this `http://<device's TS IP>:5252/ok` request is working if the browser is part of the tailnet.

But after login/authentication (to see the non-read-only UI of TS web), it opens up a *new page* with `http://<device's TS IP>:5252/?check=now`

I can't see how it can work with HA's UI.

## 3. Mixed content problem

When I access HA's ***httpS*** UI (browser -> **nginx add-on (https)** -> HA/Ingress (http) -> add-on's nginx -> tailscale web), this `http://<device's TS IP>:5252/ok` request is blocked with "(blocked:mixed-content)". As I understand the browser (Chrome) blocks the http request from a https page.

I have zero clue how to fix this. But this seems to be the side-effect of TS's home-made security/authentication code.

lmagyar:
# Proposed Changes

Until tailscale webui doesn't make available most of the add-on config options, it is better to make it readonly. If it is not readonly, changed values will be lost after the next add-on restart.

## Related Issues

closes #314

### DIFF
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -56,6 +56,11 @@ supported interfaces to Tailscale.
 Consider disabling key expiry to avoid losing connection to your Home Assistant
 device. See [Key expiry][tailscale_info_key_expiry] for more information.
 
+**Note:** _Some of the options below also available on Tailscale's web interface
+through the Web UI, but they are made read only there. You can't change them
+through the Web UI, because all the changes made there would be lost when the
+add-on is restarted._
+
 ```yaml
 accept_dns: true
 accept_routes: true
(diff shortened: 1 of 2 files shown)


frenck:
1.6.0 merged into the current main branch

lmagyar:
Tested, even with reauthentication, seems to be OK,

frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #360, opened 2024-05-19
Fix Linux optimizations

lmagyar:
# Proposed Changes

Fix optimization for configurations where there is no IPv4 default route

## Related Issues

PR #351

## Summary by CodeRabbit

- **New Features**
  - Improved network performance by optimizing UDP offloading based on the default route device.

PR #351, opened 2024-05-09
Linux optimizations for subnet routers and exit nodes

lmagyar:
# Proposed Changes

Noticed a warning in the tailscale logs:

```
Warning: UDP GRO forwarding is suboptimally configured on enu1u1, UDP forwarding throughput capability will increase with a configuration change.
See https://tailscale.com/s/ethtool-config-udp-gro
```

The suggested change at https://tailscale.com/s/ethtool-config-udp-gro basically turns on the rx-udp-gro-forwarding feature.

Smoke tested on: HA OS (armv7l, aarch64, x86_64 (VM)), Supervised (x86_64 (VM))

## Related Issues

--

PR #361, opened 2024-05-20
Allow Linux optimizations (UDP GRO for forwarding) on multiple default interfaces and IPv6

lmagyar:
# Proposed Changes

SSIA

## Related Issues

PR #351 #360

## Summary by CodeRabbit

- **New Features**
  - Enhanced network interface handling to support UDP GRO forwarding, improving network performance.

### DIFF
--- a/tailscale/rootfs/etc/s6-overlay/s6-rc.d/post-tailscaled/run
+++ b/tailscale/rootfs/etc/s6-overlay/s6-rc.d/post-tailscaled/run
@@ -4,7 +4,7 @@
 # Home Assistant Community Add-on: Tailscale
 # Runs after the machine has been logged in into the Tailscale network
 # ==============================================================================
-declare default_route_device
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

lmagyar:
And now I'm thinking why Tailscale suggested only the default interface?

That is good for exit nodes, but what about advertised subnets?

Shouldn't we do this optimization for the advertised subnets' interfaces also, like the MSS clamping?

frenck:
I cannot answer that suggestion. If they explicitly suggest it as such, there must be some kind of reason for it on their end.

lmagyar:
FYI: Certainly you are right, they check for the default interface only, for whatever reason. https://github.com/tailscale/tailscale/blob/main/ipn/ipnlocal/local.go#L5523

PR #370, opened 2024-05-22
Failsafe enabling of UDP GRO for forwarding

### LINKED ISSUE #362: Connectivity to Hass suddenly stopped - no known changes occured
# Problem/Motivation

Today I noticed I could not reach my Home assistant while external to my local and tailscale network. This configuration has been working for 4 or 5 months without issue.

## Expected behavior

Successfully access HASS through iOS app or browser from the internet

## Actual behavior

Timeout, debug logs in add on do not show the attempted connection

lmagyar:
# Proposed Changes

Revert this suggested optimization altogether, makes more problem then solves.

We could ask TS with this `curl --unix-socket /var/run/tailscale/tailscaled.sock http://local-tailscaled.sock/localapi/v0/check-udp-gro-forwarding`, but that is a bad idea, undocumented api, let them solve this on their own.

## Related Issues

fixes #362

## Summary by CodeRabbit

- **Chores**
  - Removed the `ethtool` package from the Tailscale Docker image to streamline the installation process.

- **Refactor**
  - Simplified the `post-tailscaled/run` script by removing unnecessary declarations and Linux optimizations for subnet routers and exit nodes.

### DIFF
--- a/tailscale/rootfs/etc/s6-overlay/s6-rc.d/post-tailscaled/run
+++ b/tailscale/rootfs/etc/s6-overlay/s6-rc.d/post-tailscaled/run
@@ -4,7 +4,6 @@
 # Home Assistant Community Add-on: Tailscale
 # Runs after the machine has been logged in into the Tailscale network
 # ==============================================================================
-declare interface
 declare -a options
 declare -a routes=()
 declare route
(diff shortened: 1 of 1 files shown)


lmagyar:
Or just let it silently failing?

With `ethtool -K "${interface}" rx-udp-gro-forwarding on rx-gro-list off &> /dev/null || true`???

frenck:
> Or just let it silently failing?
> 
> With `ethtool -K "${interface}" rx-udp-gro-forwarding on rx-gro-list off &> /dev/null || true`???

That is maybe not a bad idea... We could try?

lmagyar:
TS suggest this optimization conditionally (tx-udp-segmentation is enabled on their tun interface [source](https://github.com/tailscale/tailscale/blob/4214e5f71b60707699b11bfe4d5f88c5d55690e4/net/netkernelconf/netkernelconf_linux.go#L17)), the reverted code tries to do it always and on much more interface than it should (still don't understand the "no device matches name" error).

So I suggest "capturing" their warning and trying ethtool only if TS suggest it and only on the interface they suggest. They test everything, so ethtool should work when they suggest. The "capturing" fails silently (undocumented API can change), ethtool fails with warning.

Moved the code after TS is up, because we can't call TS API before it's up, so there will be a TS warning, then we set it up.

This version is quite safe to be silent on any error.

```
Warning: UDP GRO forwarding is suboptimally configured on enu1u1, UDP forwarding throughput capability will increase with a configuration change.
See https://tailscale.com/s/ethtool-config-udp-gro
[23:37:19] INFO: Tailscale is running
[23:37:20] INFO: Enable UDP GRO for forwarding on enu1u1
```

frenck (approved):
Nice piece of bash 👍 

../Frenck

Issue #372, opened 2024-05-25
Set fixed or maintain current UDP endpoint port after restart


msaario:
# Problem/Motivation

After every restart of the add-on, new random UDP port is assigned for endpoints. Under certain NAT conditions, the faster direct connections are blocked until port forwarding is set up. If I wish to allow direct connection from CG-NAT clients, I need to create new port forwarding rule after each restart of the add-on or use UPnP.

## Expected behavior

I would like to manually specify UDP port in configuration to use on endpoints or have a fixed port automatically chosen by the add-on that will stay the same after restart.

## Actual behavior

Currently new random UDP endpoint port is assigned after every restart.

## Steps to reproduce

Observe current endpoint UDP port and restart Tailscale add-on. New port is assigned randomly.

## Proposed changes

I propose optional configuration setting to keep existing chosen port or set manually UDP port for endpoint, which user can apply port forwarding rules. In Linux client this can be currently done by setting tailscaled --port 1234 or editing PORT line in /etc/default/tailscaled but these seem not to be accessible in the add-on.
As workaround, UPnP allows forwarding the randomized port automatically.

PR #390, opened 2024-07-18
Make UDP port configurable

### LINKED ISSUE #388: Ablity to specify endpoint port for easy port forwarding
# Problem/Motivation

Currently, the Docker environment variable "PORT=xxx" is not configurable, which causes the port to change on every restart of the container. This creates an issue for users who need to set a specific port for the endpoint and enable port forwarding for Tailscale without going through a DERP proxy.

## Expected behavior

There should be a configuration option to set the Docker environment variable "PORT=xxx" so that users can define a specific port. This would allow for a consistent endpoint port across reboots which would allow straightforward port forwarding.

## Actual behavior

The Docker environment variable "PORT=xxx" is configurable. This results in the endpoint port changing on every reboot of the container, complicating the setup for port forwarding and Tailscale integration, as the port must be manually adjusted each time.

## Steps to reproduce
Set up the Docker container without the ability to specify the "PORT" environment variable.
Reboot the Docker container.
Observe that the port changes on every reboot.

## Proposed changes

Add a configuration option in the Docker setup to allow users to specify the environment variable "PORT=xxx". This will enable users to set a consistent endpoint port, preventing changes on container reboots, and facilitate port forwarding for Tailscale.

lmagyar:
# Proposed Changes

Without this option TS is not always able to establish peer-to-peer connections.

See #388 for more details.

## Related Issues

fixes #388 #372

### DIFF
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -80,6 +80,7 @@ tags:
   - tag:example
   - tag:homeassistant
 taildrop: true
+udp_port: 12345
 userspace_networking: true
 ```
 
@@ -310,6 +311,17 @@ devices.
 When not set, this option is enabled by default.
 
 Received files are stored in the `/share/taildrop` directory.
+
+### Option: `udp_port`
+
+UDP port to listen on for WireGuard and peer-to-peer traffic.
+
+Use this option (and router port forwarding) if you experience that Tailscale
+can't establish peer-to-peer connections to some of your devices (usually behind
+CGNAT networks). You can test connections with `tailscale ping
+<hostname-or-ip>`.
+
+When not set, an automatically selected port is used by default.
 
 ### Option: `userspace_networking`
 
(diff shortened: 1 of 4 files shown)


tailscale/DOCS.md
```diff
@@ -311,6 +312,17 @@ When not set, this option is enabled by default.
...
 Received files are stored in the `/share/taildrop` directory.
 
+### Option: `udp_port`
+
+UDP port to listen on for WireGuard and peer-to-peer traffic.
+
+Use this option (and router port forwarding) if you experience that Tailscale
+can't establish peer-to-peer connections to some of your devices (usually behind
+CGNAT networks). You can test connections with `tailscale ping
+<hostname-or-ip>`.
```
frenck, line 322: Should we use the network ports natively supported by add-ons instead?

frenck (changes requested):

tailscale/DOCS.md
lmagyar: Yeah, you are right, this port is available on the local interfaces, not only on the tailnet, like the proxy_and_funnel_port.

The only strange user experience is, that it is hidden behind a "Show disabled ports" button, meanwhile it is not disabled but randomly selected.

PR #420, opened 2024-10-22
Make DSCP configurable on tailscaled's network traffic

lmagyar:
# Proposed Changes

It is practical if someone want's to handle TS's connection differently on his/her router. TS's packages are coming from the same IP like any other HA traffic, I couldn't find any other way to distinguish it from other HA traffic.

My use-case: only TS is allowed to use the backup WAN connection, it keeps HA's UI available, but all other traffic can be dropped until the main WAN recovers.

## Release Notes

- **New Features**
  - Introduced a new configuration option, `dscp`, allowing users to set Differentiated Services Code Point (DSCP) values for Tailscale network traffic.
  
- **Documentation**
  - Updated documentation to include the new `dscp` configuration option and clarified existing configuration instructions.

These enhancements improve user control over network traffic management, ensuring a more tailored experience with Tailscale.

### DIFF
--- a/tailscale/rootfs/etc/s6-overlay/s6-rc.d/tailscaled/run
+++ b/tailscale/rootfs/etc/s6-overlay/s6-rc.d/tailscaled/run
@@ -4,8 +4,11 @@
 # Home Assistant Community Add-on: Tailscale
 # Runs tailscale
 # ==============================================================================
+readonly TAILSCALED_GID=51892
+
 declare -a options
 declare udp_port
+declare tailscaled_gid
 
 bashio::log.info 'Starting Tailscale...'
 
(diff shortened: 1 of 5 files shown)


tailscale/rootfs/etc/s6-overlay/s6-rc.d/tailscaled/run
```diff
@@ -30,15 +33,49 @@ then
...
+    bashio::log.notice "  DSCP is already set for tailscaled"
+  elif ! ip6tables -t mangle -A OUTPUT -m owner --gid-owner ${TAILSCALED_GID} -j DSCP --set-dscp $(bashio::config "dscp"); then
+    bashio::log.warning "  Setting DSCP for tailscaled is unsuccessful"
+  fi
+fi
+
 # Run Tailscale
 if bashio::debug ; then
-  exec /opt/tailscaled "${options[@]}"
+  exec s6-setuidgid 0:${tailscaled_gid} /opt/tailscaled "${options[@]}"
```
frenck, line 72: This goes beyond what is recommend in HA. We shouldn't be running this stuff in random GIDs or UIDs

frenck (commented):
I'm going to friendly decline this PR. It adds complexity that is unneeded (and probably not used by 99.9% if the end-users). But more importantly, it is adding features not related to tailscale.

../Frenck

PR #421, opened 2024-10-22
Add HEALTHCHECK support

lmagyar:
# Proposed Changes

If Tailscale is offline longer then a configurable timeout, add-on will be restarted.

In my case a plain WAN switch-over to backup WAN on the router was able to permanently make Tailscale offline. And it's not only me: https://github.com/tailscale/tailscale/issues/12021 It seems to be an issue when TS is in a container. And similar bugs can happen in the future.

By default it won't treat offline status (.Self.Online == false) unhealthy, only Stopped BackendState (I was able to put TS into Stopped state only manually, but maybe this can be helpful once, at least won't harm).

### DIFF
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -70,6 +70,7 @@ advertise_routes:
   - 192.168.1.0/24
   - fd12:3456:abcd::/64
 funnel: false
+healthcheck_timeout: 110
 log_level: info
 login_server: "https://controlplane.tailscale.com"
 proxy: false
@@ -188,6 +189,23 @@ port 443 (or the port configured in option `proxy_and_funnel_port`)._
 
 **Note:** _If you encounter strange browser behaviour or strange error messages,
 try to clear all site related cookies, clear all browser cache, restart browser._
+
+### Option: `healthcheck_timeout`
+
+This option allows you to set timeout in seconds for Tailscale to be offline.
+
+Tailscale is quite resilient and can recover from nearly any network change. But
+in case it fails to recover and remains offline longer than healthcheck_timeout
+seconds, the add-on can be restarted. The check happens only when Tailscale is
+running, ie. it won't have any effect when Tailscale's status is eg. Starting,
+NeedsLogin or NeedsMachineAuth.
+
+The Stopped status is deemed unhealthy by default.
+
+Note: The add-on's health is checked by Home Assistant in each 30s, ie. the
+effective resolution of this option is 30s, not 1s.
+
+When not set, this option is disabled by default.
 
 ### Option: `log_level`
 
--- a/tailscale/rootfs/usr/bin/healthcheck
+++ b/tailscale/rootfs/usr/bin/healthcheck
@@ -1,0 +1,30 @@
+#!/command/with-contenv bashio
+# shellcheck shell=bash
+
+# Plain (non-json) tailscale status returns error when status is not Running or Starting, so eg. NeedsLogin and NeedsMachineAuth would make it unhealthy
+# The .Health json filter returns any problems, so even temporary health problems would make it unhealthy
+# This script treats the following situations unhealthy:
+# - always: .BackendState == "Stopped"
+# - optionally: .BackendState == "Running" && .Self.Online == false for more then healthcheck_timeout seconds (configurable)
+#   This can handle internal TS bugs, like https://github.com/tailscale/tailscale/issues/12021 where TS fails to recover from rerouting traffic from normal WAN to failover WAN
+
+declare status_json
+declare backend_state self_online
+# LAST_ONLINE_TIMESTAMP is in contenv at /var/run/s6/container_environment
+
+status_json=$(/opt/tailscale status --json --self=true --peers=false)
+backend_state=$(jq -r '.BackendState' <<< "${status_json}")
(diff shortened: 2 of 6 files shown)


lmagyar:
This behaves as expected in live environments and no issues reported by real users.

tailscale/DOCS.md
```diff
@@ -70,6 +70,8 @@ advertise_routes:
   - 192.168.1.0/24
   - fd12:3456:abcd::/64
 funnel: false
+healthcheck_offline_timeout: 110
+healthcheck_restart_timeout: 3600
```
frenck, line 74: I really don't want to keep add options like this.

People shouldn't have to worry about this stuff.

../Frenck

frenck (changes requested):

tailscale/DOCS.md
lmagyar: Hardwire 5 minutes and 1 hour? Based on my logs, 2-3 minutes long dropouts occure approx. daily.

It also means that healthcheck will be live after update. I'm just a bit afraid that a "regular" restart can ruin something. Maybe DNS. I was never able to configure HA to use TS's DNS, so I can't test it, it's just a wild guess.

lmagyar:
@coderabbitai review

tailscale/rootfs/usr/bin/healthcheck
```diff
@@ -0,0 +1,60 @@
+#!/command/with-contenv bashio
+# shellcheck shell=bash
+
+# Plain (non-json) tailscale status returns error when status is not Running or Starting, so eg. NeedsLogin and NeedsMachineAuth would make it unhealthy
+# The .Health json filter returns any problems, so even temporary health problems would make it unhealthy
+# This script treats the following situations unhealthy:
+# - .BackendState == "Stopped"
+# - if once was online but gets offline for more then HEALTHCHECK_OFFLINE_TIMEOUT seconds
+# - if never gets online for more then HEALTHCHECK_RESTART_TIMEOUT seconds
+#   This can handle internal TS bugs, like https://github.com/tailscale/tailscale/issues/12021 where TS fails to recover from rerouting traffic from normal WAN to failover WAN
```
frenck, line 10: We shouldn't be handling upstream bugs. That should be fixed.
The fact this is here isn't correct.

Masking issues will never help solving them.

tailscale/rootfs/usr/bin/healthcheck
```diff
@@ -0,0 +1,60 @@
...
+  bashio::log.fatal "Add-on is unhealthy"
+  bashio::exit.nok
+else
+  # Healthy
+  if ! bashio::var.equals "${LAST_REPORTED_HEALTH_STATE-}" "HEALTHY"; then
+    printf "HEALTHY" > /var/run/s6/container_environment/LAST_REPORTED_HEALTH_STATE
+    # Log it only once
+    bashio::log.info "Add-on is healthy"
+  fi
+fi
```
frenck, line 60: This seems unneeded. Let's remove it.

tailscale/rootfs/usr/bin/healthcheck
```diff
@@ -0,0 +1,60 @@
...
+    (( $(date +"%s") - ${LAST_ONLINE_TIMESTAMP} > ${HEALTHCHECK_OFFLINE_TIMEOUT} )) ) || \
+  (! bashio::var.has_value "${LAST_ONLINE_TIMESTAMP-}" && \
+    (( $(date +"%s") - ${STARTED_TIMESTAMP} > ${HEALTHCHECK_RESTART_TIMEOUT} )) )
+then
+  # Unhealthy
+  if ! bashio::var.equals "${LAST_REPORTED_HEALTH_STATE-}" "UNHEALTHY"; then
+    printf "UNHEALTHY" > /var/run/s6/container_environment/LAST_REPORTED_HEALTH_STATE
+  fi
+  # Log it always
+  bashio::log.fatal "Add-on is unhealthy"
```
frenck, line 51: Let's log: Why

frenck (changes requested):

tailscale/DOCS.md
```diff
@@ -347,6 +347,19 @@ CGNAT networks). You can test connections with `tailscale ping
...
 When not set, an automatically selected port is used by default.
 
+## Healthcheck
+
+Tailscale is quite resilient and can recover from nearly any network change. In
+case it fails to recover, the add-on's health is set unhealthy. The add-on's
+health is checked by Home Assistant in each 30s, and if it reports itself 3
+times unhealthy in a row, the add-on will be restarted.
+
+The add-on's health is set unhealthy:
```
frenck, line 357: Healthy or not, and even the healthcheck itself, is an internal; not user-facing.

Explaining this to an end-user seems wrong, as they have no clue what an "health check" even remotely it.

This part of the docs would thus be fully up in the air for most readers.

tailscale/rootfs/usr/bin/healthcheck
lmagyar: Removed the reference to the issue.

lmagyar:
Removed the log and the docs, now this is always workign in the background.

lmagyar:
not stale

lmagyar:
not stale

frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

Issue #439, opened 2024-12-24
Update or add current Tailscale terminology


KlfJoat:
# Problem/Motivation

Similar to home-assistant/core#132976, it would probably be a good idea to align the Add-On's terminology, especially documentation, with that of Tailscale.

## Expected behavior

tailscale/DOCS.md to mention somewhere that "proxy" = Tailscale serve.

And maybe for the "proxy" YAML item to be named "serve".

## Actual behavior

The word "serve" is nowhere on the page, nor is there a link to the Tailscale serve documentation. It links to the HTTPS stuff, and that's good.

## Proposed changes

Either rename the "proxy" option to "serve", add the word "serve" or an explanation that "proxy" = "serve" somewhere.

Also, especially add a link to the Tailscale serve documentation. https://tailscale.com/kb/1312/serve

I understand that for novice users, the term "proxy" may be desirable because that's what Tailscale serve does. I also recognize that I don't know how easy it would be to change the "proxy" item in the YAML config. So I'm not going to suggest that those get removed entirely. But for those of us who know something about Tailscale, mentioning "serve" somewhere would be helpful.

For me, it was only the repeated alignment of "proxy" with "funnel", and the vague recollection I have that "serve" might have had another name in the past that made me try it out. The logs confirmed it was really Tailscale serve.

lmagyar:
Thanks for the feedback, you are right, things has changed.

Adding the missing link is a no-brainer.

Originally both the reverse proxy and the funnel was an option for the serve command. Additionally the current 2 separate bool config option (proxy and funnel) is also, hmmm, not so good. See also #407. So I suggest to merge them, but I really struggle with a good naming (it's harder than developing this stuff). What do you think about this:

config.yaml:
```diff
-  funnel: bool?
-  proxy: bool?
+  share: list(disabled|tailnet|tailnet and internet)?
-  proxy_and_funnel_port: match(^(443|8443|10000)$)?
+  share_port: match(^(443|8443|10000)$)?
```

And the documentation can explain/reference serve for the tailnet option and funnel for the internet option.

Tailscale uses similar wording, like (this is a funnel log):
```
Available on the internet:
https://xxxx.tailxxxx.ts.net/
|-- proxy http://127.0.0.1:xxxx
```
And uses similar wording in the online serve/funnel documentation "You can think of this as sharing the service, ... with the rest of your tailnet." "You can use it to share a local service,"

[...]

PR #445, opened 2025-01-02
Merge proxy and funnel options into share_homeassistant, rename proxy_and_funnel_port to share_on_port (nonbreaking change, config is automatically updated)

lmagyar:
# Proposed Changes

***Note: This is not a breaking change, this updates the config automatically on startup!***

The standalone, dual, bool? `proxy` and `funnel` options were wrong on many levels:
- they weren't independent, docs said: enable proxy if funnel is enabled
- naming was 'misleading': proxy vs serve
- continuous requests to extend serve and funnel with something

After several naming attempt, this seems to be the most logical to me:
- it says: share HA with serve/funnel, ie. this is not a general serve/funnel config option
- the DOCS.md also got shorter and more logical

Screenshot:

image

## Related Issues

fixes: #439 (and several closed old ones)

## Summary by CodeRabbit

- **Documentation**
  - Consolidated Tailscale add-on docs to use new options and clarifies Serve vs Funnel usage and TLS handling.

- **New Features**
  - Introduced unified sharing options enabling Serve or Funnel sharing with valid certificates and improved validation/messages.

- **Configuration Changes**
  - Replaced `funnel`, `proxy`, and `proxy_and_funnel_port` with `share_homeassistant` and `share_on_port`; default port 443.

- **Chores**
  - Added migration/upgrade logic to convert and clean up deprecated options automatically.


tailscale/config.yaml
sinclairpaul, line 1: Removing the Schema options would cause validation errors for existing users wouldn't it?, I know we added a function to bashio to remove schema options, we could possibly add something to add and remove?

lmagyar: It works fine. Old options remain in the config, accessible with bashio::addon.options (and updateable without any constraint with bashio::addon.option), only bashio::addon.config returns the "filtered" config, where only the valid options are available.

This is a quite straightforward concept IMHO to allow add-on devs to update config options on startup.

And the old options can be seen in the yaml UI view also. And the yaml config still allowed to save, they are out of the schema validation, that validates only the "current/new" options even in the UI. This seems to be also a correct approach, they are not deleted automatically, and users are not forced to remove them.

sinclairpaul: So no errors on save with the old options, including the supervisor logs?

sinclairpaul: Also from what we added previously, I believe you can do it via `bashio::addon.option()`

lmagyar: Wow, you are right, there are SU warnings: WARNING (MainThread) [supervisor.addons.options] Option 'xxx' does not exist in the schema for xxx

But I can edit it in the UI, it is saved, checked with bashio::addon.options, it just works. You can even add foo: bar as a new option in the yaml editor. :/

Yes, I already use bashio::addon.option. :)

sinclairpaul: Yeah the validation isn't the best, however I would suggest we read the option and replace it with if exists via the function.  Will save the issues six months down the line with users with auto update...

lmagyar: Yes, the upgrade is already done at the beginning of [tailscale/rootfs/etc/s6-overlay/scripts/stage2_hook.sh](https://github.com/hassio-addons/addon-tailscale/pull/445/files#diff-dc7259da1e04caadd09f3f46bfd4c75733f46d909343e1d8f41d0bbed2160d01R11)

That's why I wrote, that this is not a breaking change.

sinclairpaul: Apologies, I missed it, ignore my comments on it :)

I would still be inclined to leave the old schema entries as optional, and remove them when the migration script is removed.

lmagyar: Hmmm, what would be the use case for the old options in the schema? That would mean 2 alternative way for the same configuration, I don't think this is a good idea. Even after this PR, if somebody configures the old options, they will be migrated to the new form on the next startup.

Though the current migration would migrate "brute forced" entries, like proxy_and_funnel_port: 123, that is not a valid value, but out of the new schema, no validation in the UI, and the current migration script doesn't validate them. I prefer to add some validation in the migration:
- validate the port value - the bool options are checked whether equal to "true", that is a clean case
- drop old options if the new options are already configured, and log a warning

lmagyar: Finally pushed some changes, this survives everything:
- if user edits/adds back the old config values, but the new values are already configured -> warning, old values are dropped
- if user edits/adds back the old config value for proxy_and_funnel_port with invalid value to upgrade it to share_on_port (UI yaml editor won't validate old option values against the new schema), and share_on_port doesn't exists, so should be upgraded -> warning, old invalid value is dropped

Note: As of S6 overlay 3.2.0.2, we can't fail in stage2_hook (and I think this won't be a good idea for this case, when invalid old config values "injected" with the yaml config editor), this is possible only in 3.2.0.3 (https://github.com/just-containers/s6-overlay/commit/df0599220f9f1552be131e52a237b9769f48fe97)

tailscale/rootfs/etc/s6-overlay/scripts/stage2_hook.sh
```diff
@@ -5,6 +5,61 @@
...
+declare share_homeassistant share_on_port
+
+# This is to execute potentially failing supervisor api functions within conditions,
+# where set -e is not propagated inside the function and bashio relies on set -e for api error handling
+function try {
+    set +e
+    (set -e; $@)
+    declare -gx TRY_ERROR=$?
+    set -e
+}
```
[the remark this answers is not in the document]
lmagyar: done

lmagyar:
Marking it draft: I will convert this logic into a oneshot service after banner and log-level, this is a bit ugly to see a log message like 'INFO: Successfully migrated proxy and funnel options to share_homeassistant: funnel' before the main add-on banner.

lmagyar:
No change, ready for review, stage2_hook uses the new options in conditions, they have to be there, so config upgrade must happen at the beginning of stage2_hook.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-homeassistant/run
```diff
@@ -44,23 +43,20 @@ else
...
     bashio::log.error "Tailscale's Funnel support is disabled"
     bashio::exit.nok
   fi
-  tailscale_command=funnel
 fi
 
-# Set up serve
-exec /opt/tailscale ${tailscale_command} --https=$(bashio::config "proxy_and_funnel_port" "443") --set-path=/ "http://127.0.0.1:$(bashio::core.port)"
+# Set up serve or funnel
+exec /opt/tailscale $(bashio::config 'share_homeassistant') --https=$(bashio::config 'share_on_port' '443') --set-path=/ "http://127.0.0.1:$(bashio::core.port)"
```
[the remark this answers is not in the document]
lmagyar: done

lmagyar:
@coderabbitai review

lmagyar:
not stale

lmagyar:
still not stale

tailscale/rootfs/etc/s6-overlay/scripts/stage2_hook.sh
```diff
@@ -5,6 +5,60 @@
 # S6 Overlay stage2 hook to customize services
 # ==============================================================================
 
+declare options
+declare proxy funnel proxy_and_funnel_port
+declare share_homeassistant share_on_port
+
```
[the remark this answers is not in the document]
lmagyar: done

tailscale/rootfs/etc/s6-overlay/scripts/stage2_hook.sh
```diff
@@ -5,6 +5,60 @@
...
+# Upgrade to share_homeassistant
+if bashio::var.true "${proxy}"; then
+    if bashio::var.true "${funnel}"; then
+        bashio::addon.option 'share_homeassistant' 'funnel'
+        bashio::log.info "Successfully migrated proxy and funnel options to share_homeassistant: funnel"
+    else
+        bashio::addon.option 'share_homeassistant' 'serve'
+        bashio::log.info "Successfully migrated proxy and funnel options to share_homeassistant: serve"
+    fi
+fi
```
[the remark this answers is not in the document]
lmagyar: nope, mandatory option, filled with default value by the supervisor

tailscale/rootfs/etc/s6-overlay/scripts/stage2_hook.sh
```diff
@@ -5,6 +5,60 @@
...
+# Upgrade to share_on_port
+if bashio::var.has_value "${proxy_and_funnel_port}"; then
+    try bashio::addon.option 'share_on_port' "^${proxy_and_funnel_port}"
+    if ((TRY_ERROR)); then
+        bashio::log.warning "The proxy_and_funnel_port option value '${proxy_and_funnel_port}' is invalid, proxy_and_funnel_port option is dropped, using default port."
+    else
+        bashio::log.info "Successfully migrated proxy_and_funnel_port option to share_on_port: ${proxy_and_funnel_port}"
+    fi
+fi
```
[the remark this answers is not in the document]
lmagyar: see previous

PR #450, opened 2025-01-13
Update documentation about DNS configuration

lmagyar:
# Proposed Changes

There are several issues about DNS settings, this doc update tries to clarify the basic configs.

It's a bit tricky, because TS add-on runs in a container and some of the TS docs and forum best practices doesn't fit perfectly for this situation.

## Related Issues

discussion #449


lmagyar:
MagicDNS conflict with hassio_dns has to be resolved first.

PR #463, opened 2025-02-05
Update documentation about UI bug in configuration editor - or make all the config options mandatory

### LINKED ISSUE #462: userspace_networking needs to be enabled and disabled to allow for peer to peer connections
# Desc
I setup tailscale for a peer to peer connection between to HA devices to be able to use Remote HomeAssistant. 

Even when seeing both instance connected in the tailscale dashboard, I was not able to ping the internal IPs from within the HA terminal.
I needed to enable and then disable the `userspace_networking`.

See post where I found this: https://community.home-assistant.io/t/remote-access-to-two-ha-instances/424162/19

After I've done this, I can ping the other instance directly. Final configuration:

Image

I also think the documentation would benefit from a section on how to connect two different HA instance with the settings from above. 

# State
Tailscale version: 0.24.0 

Image

lmagyar:
# Proposed Changes

This issue is much more serious than it seems, users always show configurations where most of the default enabled options are enabled manually, and complain that disabled options are not working or hard to make them to work.

They waste hours on forums and debugging a configuration issue caused by a very misleading UI config editor!

## Related Issues

fixes #462

## Summary by CodeRabbit

- **Documentation**
	- Enhanced guidance for configuring the Tailscale add-on.
	- Added cautionary notes to help users avoid configuration pitfalls with the Home Assistant UI.
		- Advises against using the "Show unused optional configuration options" switch to prevent confusion.
		- Emphasizes the necessity of manually updating optional settings via YAML for accurate configuration.
		- Warns that the UI may misrepresent the state of optional configuration options, leading to potential misunderstandings.

### DIFF
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -89,17 +89,17 @@ userspace_networking: true
 > optional configuration options" switch on the Configuration tab!
 
 > [!CAUTION]
-> When you want to change the default behaviour of any of these optional
-> configuration options, **add them to the YAML add-on configuration manually**,
-> by using the "Edit in YAML" in the ... menu on the right and save them! Use
-> the UI to edit them only after you added them manually!
+> When you want to change the default behaviour of these optional configuration
+> options, **add them to the YAML add-on configuration manually**, by using the
+> "Edit in YAML" in the ... menu on the right and save them! Use the UI to edit
+> them only after you added them manually!
 
 > [!WARNING]
 > Home Assistant's UI will show you all the optional configuration options
-> turned off instead of grayed out, and you will falsely believe that those are
-> the values that will be stored, so you will enable options, that by default
-> are already enabled when unused, and you will let options be disabled, that
-> you originally wanted to disable, but the **UI will not add these disabled
+> turned off instead of grayed out. You will falsely believe that those are the
+> values that will be stored, so you will enable options, that by default are
+> already enabled when unused, and you will let options be disabled, that you
+> originally wanted to disable. But the **UI will not add these disabled
 > optional configuration options** to the YAML add-on configuration, and at the
 > end nothing will change in the add-on's functionality after a restart!
 
--- a/tailscale/DOCS.md
+++ b/tailscale/DOCS.md
@@ -84,6 +84,25 @@ userspace_networking: true
 > change them through the Web UI, because all the changes made there would be
 > lost when the add-on is restarted.
 
+> [!CAUTION]
+> Due to limitations in Home Assistant's UI, **do not use** the "Show unused
+> optional configuration options" switch on the Configuration tab!
+
+> [!CAUTION]
+> When you want to change the default behaviour of any of these optional
+> configuration options, **add them to the YAML add-on configuration manually**,
+> by using the "Edit in YAML" in the ... menu on the right and save them! Use
+> the UI to edit them only after you added them manually!
+
+> [!WARNING]
+> Home Assistant's UI will show you all the optional configuration options
+> turned off instead of grayed out, and you will falsely believe that those are
+> the values that will be stored, so you will enable options, that by default
+> are already enabled when unused, and you will let options be disabled, that
+> you originally wanted to disable, but the **UI will not add these disabled
+> optional configuration options** to the YAML add-on configuration, and at the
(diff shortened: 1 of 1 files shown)


tailscale/DOCS.md
```diff
@@ -83,6 +83,25 @@ taildrop: true
 userspace_networking: true
 ```
 
+> [!CAUTION]
+> Due to limitations in Home Assistant's UI, **do not use** the "Show unused
+> optional configuration options" switch on the Configuration tab!
```
frenck, line 88: This is not something I want to document.

It is a problem upstream.

frenck (changes requested):

tailscale/DOCS.md
lmagyar: How else warn users who spend hours to figure out, that the UI does the exact opposite what it shows?

Thats why nearly all the yaml configs users share show they enable the already by-default-enabled options, and they notice only after hours that the UI never saves the seemingly disabled options, the options they originally wanted to disable when they opened the config tab.

lmagyar: There is an open RFC about add-on config options, this is opened 4 years ago, nothing happened. https://github.com/home-assistant/supervisor/issues/2640

Meanwhile de facto issues/bugfixes are closed, that it is better to implement the RFC. https://github.com/home-assistant/supervisor/issues/4606 https://github.com/home-assistant/supervisor/pull/4607

Even your issue isn't fixed correctly. https://github.com/home-assistant/frontend/issues/8614

If even you can't get something fixed by the frontend team, it will never be fixed.

17.000+ users of this add-on, fighting for hours to configure even the simplest thing, they deserve a warning about this issue.

When it get fixed, I will remove this warning from the docs.

ggsp: Complete newbie to Home Assistant here so take my opinion with a handful of salt—I was trying to figure out how to stop my HAOS instance from trying to advertise itself as an exit node, and was able to figure out how only thanks to this PR. Having this kind of warning front and center would've sped things up!

frenck: Sorry, I'm not adding this.

Working around issues will never fix it.

frenck: > Complete newbie to Home Assistant here so take my opinion with a handful of salt—I was trying to figure out how to stop my HAOS instance from trying to advertise itself as an exit node, and was able to figure out how only thanks to this PR. Having this kind of warning front and center would've sped things up!

Please report the issue upstream. Learning to workaround the problem will help no-one.

../Frenck

frenck:
Thinking about this more, let's remove the optional flags.

That should take care of everything. If this is problematic, lets just make everything mandatory.

lmagyar:
> make everything mandatory

1 option is problematic: `advertise_routes`

Where:
- undefined = use routes from the HA managed/supported interfaces
- empty list [] = disabled

Here I can't figure out a nice way to configure the current default behaviour, ie. using the supported interfaces. The `advertise_routes: ~` or `advertise_routes: null` is not accepted by the UI.

Converting it into an empty list, would break existing configs.

Using:
```
options:
  # advertise_routes: # intentionally left empty
schema:
  advertise_routes:
    - "match(^...$)"
```
And populating it on startup can work, but fails if user tries to edit the config before starting the add-on up.

I run out of ideas.

lmagyar:
not stale

lmagyar:
Thinking about this, maybe this can work:

```
options:
  advertise_routes:
    - local subnets
schema:
  advertise_routes:
    - "match(^(local.subnets|...)$)"
```

But what to do, if users configure this? We can merge/remove duplicate entries, seems reasonable, and already implemented in the scripts:
```
options:
  advertise_routes:
    - local subnets
    - 192.168.1.0/24
```

tailscale/DOCS.md
ananthb: > Complete newbie to Home Assistant here so take my opinion with a handful of salt—I was trying to figure out how to stop my HAOS instance from trying to advertise itself as an exit node, and was able to figure out how only thanks to this PR. Having this kind of warning front and center would've sped things up!

Documenting this bug has helped at least one person so far.

ananthb:
Not stale.

lmagyar:
Closing in favor of #541.

PR #506, opened 2025-05-15
Remove duplicated service dependencies

lmagyar:
# Proposed Changes

These dependencies from the user bundle service are unnecessary:
- init-nginx and web always started by nginx
- protect-subnets started by post-tailscaled on demand (stage2-hook removes the dependency if not needed)
- tailscaled is always started by post-tailscaled

S6 is clever enough to figure out what should be started, we should add only "user" facing services to the user bundle.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of configuration conditions to prevent unintended removal of certain system files.


frenck (approved):
Thanks, @lmagyar 👍 

../Frenck

PR #624, opened 2026-02-05
Make the proxied internal port configurable

rayanamal:
# Proposed Changes

This simple change makes the proxied internal port configurable by adding a `share_from_port` option. This allows for sharing other apps than just Home Assistant (e.g. Immich) which are served from a different internal port.

The other proxy service that is popular within the community, Cloudflare Tunnels, can serve from a different internal port. But, for many apps (add-ons) it's simply not suitable, because it's subject to rather low bandwidth limits (anecdotally ~10Mbps).

While Tailscale Funnel has similar bandwidth limits, unlike Cloudflare Tunnels, it actually works extremely well for serving both high and low bandwidth use cases, because:

- When a client device is not connected to the VPN, Funnel allows sufficient bandwidth over Funnel for serving web apps (like Home Assistant) or basic low-bandwidth operations like background photo/file sync.
- When a client device is connected to the VPN, it allows near unlimited bandwidth (e.g. for video or movie streaming, file upload/download, etc.).

Most importantly, this switch happens automatically at the network level, so client applications (e.g. Immich) don't even need to handle, nor are they aware of this switch. The same access URL is used in both cases. When connected to the VPN, Tailscale transparently switches the connection from the public proxied Funnel connection to the p2p Wireguard connection.

It's well known that leaving the Tailscale VPN always connected on your mobile device drains your battery. Thankfully, with this setup, it's possible to connect to the VPN only when needed (e.g. when you open immich, Plex, Jellyfin, Nextcloud, etc. mobile apps) and let background sync and other low-bandwidth operations continue to work even when not connected. It's trivial too to automate connecting to VPN when an app is opened on iOS and Android.

This makes Tailscale Funnel an extremely powerful choice for many such self-hosted applications. Basically, there is no substitute for it today.

The default is preserved as the Home Assistant frontend port (8123), so existing and future users will not be affected unless they explicitly set this option.

## Related Issues

> https://github.com/hassio-addons/app-tailscale/issues/431

## Summary by CodeRabbit

* **New Features**
  * Added an option to specify which internal port is exposed externally (default: 8123).
* **Documentation**
  * Updated docs, README and translations with the new option and revised installation/FAQ content.
* **Bug Fixes / Stability**
  * Adjusted internal web/backend ports and startup readiness checks for improved reliability.
* **Chores**
  * Updated bundled web server package version.

### DIFF
--- a/tailscale/Dockerfile
+++ b/tailscale/Dockerfile
@@ -16,7 +16,7 @@ RUN \
         ipcalc=1.0.3-r1 \
         iproute2=6.17.0-r0 \
         iptables=1.8.11-r1 \
-        nginx=1.28.0-r8 \
+        nginx=1.28.2-r0 \
         coreutils=9.8-r1 \
         networkmanager=1.54.3-r0 \
     \


rayanamal:
I understand the reason behind not accepting this, closing.

PR #684, opened 2026-05-08
Use release drafter as standalone workflow again

lmagyar:
This is draft, because it waits for a new workflow repo release!

# Proposed Changes

After https://github.com/hassio-addons/workflows/pull/291 it reverts #638 and #639.


lmagyar:
Not stale, we are waiting for a new release of https://github.com/hassio-addons/workflows.

lmagyar:
Merged into #700.

PR #702, opened 2026-06-16
Allow serving / funneling even if HA is set up with HTTPS/SSL

dynamyc010:
# Proposed Changes

Ever since I switched my Home Assistant instance over to my own homelab SSL, it caused Serve and Funnel to not function (as they really want to use HTTP over HTTPS). 

Since Tailscale supports serving HTTPS even if it doesn't trust the cert (with HTTPS+Insecure), I just refactored the run script to add support for it.

## Related Issues

I couldn't find any issues for it, but since it was such a small thing relatively, I just did it myself in like an hour.

## Testing

I tested in my own environment, Tailscale starts up properly and the Funnel is served properly as well now.

## Summary by CodeRabbit

* **Bug Fixes**
  * Home Assistant connections now adapt to HTTP or HTTPS configurations, improving startup and proxy connectivity.
  * Secure connections with self-signed certificates are supported.
  * Funnel capability is validated before Home Assistant connection setup.
* **Documentation**
  * Updated the “share_homeassistant” setup guidance to clarify HTTPS support through Tailscale and recommend using a reverse proxy for separate HTTPS access.


lmagyar:
Have you tested this before submitting a PR? Because it doesn't work.

Execute in the app's container `/opt/tailscale serve --bg=false --https=10000 --set-path=/ https+insecure://127.0.0.1:xxxx` Results in 502, and funnel is the same. I've tested it in my fork years ago, and it didn't work then, and removed it (https://github.com/lmagyar/homeassistant-addon-tailscale/commit/33d2d71d748b603e0724500601f0b3b07bceae3d). As far as I remember the proxy that implements serve doesn't accept the cert of your HA on 127.0.0.1.

You have to run HA on http and use NGINX proxy app to proxy it through https on your non-TS connections.

dynamyc010:
I am currently running this branch as a local version, and it boots fine, and I can access it on the outside internet without Tailscale as well. 
image
(whoops, the blurring is terrible, but I give up)

If I try and start up the current release, it just keeps spamming about the HTTPS unsupported error.

The actual cert given to Home Assistant is a self-CA cert, so there's no way Tailscale would otherwise like it.
image

dynamyc010:
One configuration change I did have to change was telling the Funnel to use 8443 instead of 443, as something seems to take up 443 somewhere, and I didn't debug that yet. Outside access still works without using a port just fine.

lmagyar:
OK, you are right!

I tested above with NGINX https->http proxy in the queue, that proxy answered 502 for the TS proxy, but when I add
```
http:
  ...
  ssl_certificate: /ssl/fullchain.pem
  ssl_key: /ssl/privkey.pem
```
to HA config, then everything works (after some browser cache, etc. cleanup).

Hmmm, I remember it failed with some certificate error 2 years ago, it seems TS fixed something.

When I have time, I will check what is needed in NGINX to make it work (OK, it's unrelated), and make some changes to this PR (it's easier to do than explain, some docs, some refactoring, you will see), I mark it draft until.

ilanbiala:
Seems like a nice feature! What is needed to land this?

frenck (approved):
Thanks, @dynamyc010 👍 

../Frenck  

<p>
  <a href="https://www.linkedin.com/in/frenck/"></a>
  &nbsp;&nbsp;
  <a href="https://youtube.com/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://bsky.app/profile/frenck.social"></a>
  &nbsp;&nbsp;
  <a href="https://fosstodon.org/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.instagram.com/frenck/"></a>
  &nbsp;&nbsp;
  <a href="https://www.threads.net/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.facebook.com/frenck.dev/"></a>
  &nbsp;&nbsp;
  <a href="https://x.com/frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.tiktok.com/@frenck.nl"></a>
</p>

Blogging my personal ramblings at <a href="https://frenck.dev">frenck.dev</a>

PR #715, opened 2026-07-17
Add support for advertising Tailscale Services

apfohl:
# Proposed Changes

This adds a new `services` configuration option that allows users to advertise local services running on the Home Assistant host as Tailscale Services.

A new S6 oneshot service `services` reads the configured services and calls `tailscale serve --service=svc:<name>` for each entry (and also removes unused/stale configurations). The `svc:` prefix needs to be added in the config manually, like the `tag:` prefix for advertised_tags.

Supported protocols are `http`, `https`, `tcp`, and `tls-terminated-tcp`. Services must be defined and approved in the Tailscale admin console before they can accept traffic.

Also updates the app documentation and translations for the new option.

Additionally on first start it imports any previously manually configured (with tailscale cli) service configurations into the app config.

## Related Issues

None. As far as I have looked into the issues.

## Summary by CodeRabbit

* **New Features**
  * Added support for exposing local applications as Tailscale Services through configuration.
  * Supports configurable service names, targets, protocols, ports, and optional URL paths.
  * Added validation and startup handling for configured services, including HTTPS requirements.
  * Added documentation and localized configuration guidance, including admin-console approval and disabled-by-default behavior.


tailscale/config.yaml
```diff
@@ -78,6 +79,12 @@ schema:
   login_server: url
   share_homeassistant: list(disabled|serve|funnel)
   share_on_port: match(^(?:443|8443|10000)$)
+  services:
+    - name: "match(^[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$)"
+      target: "match(^(?:https?|tcp)://127\\.0\\.0\\.1:\\d+$)"
```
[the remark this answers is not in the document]
apfohl: fixed

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
+#!/command/with-contenv bashio
+# shellcheck shell=bash
+export LOG_FD
```
lmagyar: This line is not needed with the new base image with the new bashio. Will be removed in all other places later, with all the temporary workarounds.

tailscale/DOCS.md
```diff
@@ -317,6 +318,47 @@ Only ports 443, 8443, and 10000 are allowed by Tailscale.
 
 Port 443 is used by default.
 
+### Option: `services`
+
+This option allows you to advertise other local services running on this device
+as [Tailscale Services][tailscale_info_services]. Each service needs a name, a
```
lmagyar: At all other places in DOCS the reference to the TS docs is in a separate line as "More information: ..." This is "important", because the device must to use tags!

tailscale/DOCS.md
```diff
@@ -317,6 +318,47 @@ Only ports 443, 8443, and 10000 are allowed by Tailscale.
...
+local target address, a protocol, and a port to expose it on. The `svc:` prefix
```
lmagyar: The advertise_tags option requires the tag: prefix. Maybe we should require svc: here also. What do you think?

tailscale/DOCS.md
```diff
@@ -317,6 +318,47 @@ Only ports 443, 8443, and 10000 are allowed by Tailscale.
...
+is added automatically by the app; do not include it in the configuration.
+
+You can use this to expose apps running on your Home Assistant instance, such as
+an audiobookshelf add-on, to your tailnet using a stable MagicDNS name.
```
lmagyar: no add-ons anymore, use apps please

tailscale/config.yaml
```diff
@@ -78,6 +79,12 @@ schema:
   login_server: url
   share_homeassistant: list(disabled|serve|funnel)
   share_on_port: match(^(?:443|8443|10000)$)
+  services:
+    - name: "match(^[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$)"
+      target: "match(^(?:https?|tcp)://127\\.0\\.0\\.1:([1-9]\\d{0,3}|[1-5]\\d{4}|6[0-4]\\d{3}|65[0-4]\\d{2}|655[0-2]\\d|6553[0-5])$)"
+      protocol: "list(http|https|tcp|tls-terminated-tcp)"
+      port: "match(^([1-9]\\d{0,3}|[1-5]\\d{4}|6[0-4]\\d{3}|65[0-4]\\d{2}|655[0-2]\\d|6553[0-5])$)"
```
lmagyar: there is a "port" type, see https://developers.home-assistant.io/docs/apps/configuration#options--schema

tailscale/rootfs/etc/s6-overlay/scripts/stage2_hook.sh
```diff
@@ -133,3 +133,8 @@ fi
 if bashio::config.equals 'share_homeassistant' 'disabled'; then
     rm /etc/s6-overlay/s6-rc.d/user/contents.d/share-homeassistant
 fi
+
+# Disable share-services service when no services are configured
```
lmagyar: nope, it always has to start, to be able to remove unused services!!!

tailscale/rootfs/etc/s6-overlay/scripts/stage2_hook.sh
```diff
@@ -133,3 +133,8 @@ fi
...
+if ! bashio::config 'services' | jq -e '. | length > 0' > /dev/null; then
+    rm -f /etc/s6-overlay/s6-rc.d/user/contents.d/share-services
```
lmagyar: don't follow every advice from an AI, you don't need -f, if it is not there, fail during development :)

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+declare target
+
+declare path
+
+declare -a options=()
+declare -a configured_services=()
+declare -a configured_names=()
+declare -a current_names=()
+
+# Nothing to do when no services are configured
```
lmagyar: nope, we still have to remove removed services!!!

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+fi
+
+# Read configured services
+# Note: bashio::config emits one JSON object per line for array options
+readarray -t configured_services < <(bashio::config 'services')
+
+# Build a list of configured service names (without the svc: prefix)
+for service in "${configured_services[@]}"; do
+    name=$(jq -re '.name' <<< "${service}")
+    name="${name#svc:}"
```
lmagyar: do we have svc: prefix there?

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+    configured_names+=("${name}")
+done
+
+bashio::log.info "Configured service hosts: ${configured_names[*]}"
+
+# Wait for tailscaled to be available
```
lmagyar: post-tailscaled already waited for this

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+while ! bashio::fs.socket_exists "/var/run/tailscale/tailscaled.sock"; do
+    sleep 1
+done
+
+while ! /opt/tailscale status --json --peers=false --self=false \
```
lmagyar: post-tailscaled already waited for this

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+    sleep 2
+done
+
+# Read currently advertised services from the tailscale serve config
+if ! current_config=$(/opt/tailscale serve get-config --all 2>/dev/null); then
+    current_config='{}'
+fi
+
+# Build a list of currently configured service names
+readarray -t current_names < <(jq -re '.services | keys[]' <<< "${current_config}" 2>/dev/null || true)
```
lmagyar: why not use bashio::jq() ?
-e and dev/null and ||true, we don't need this, use bashio::jq IMHO

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+
+# Remove services that are no longer configured in this app
+for service_name in "${current_names[@]}"; do
+    if [[ "${service_name}" != svc:* ]]; then
```
lmagyar: why do wee need this? just asking! what else can be there?

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+        continue
+    fi
+    # Compare without the svc: prefix
+    if ! printf '%s\n' "${configured_names[@]}" | grep -qx "${service_name#svc:}"; then
+        bashio::log.info "Removing stale service host configuration for ${service_name}"
+        /opt/tailscale serve drain "${service_name}" >/dev/null 2>&1 || true
```
lmagyar: we should log at least a warning on error (an exit.nok would stop the app, so let it start up, but don't hide the error)

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+    exit 0
+fi
+
+# Read configured services
+# Note: bashio::config emits one JSON object per line for array options
+readarray -t configured_services < <(bashio::config 'services')
+
+# Build a list of configured service names (without the svc: prefix)
+for service in "${configured_services[@]}"; do
+    name=$(jq -re '.name' <<< "${service}")
```
lmagyar: bashio:jq ?

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+    if ! printf '%s\n' "${configured_names[@]}" | grep -qx "${service_name#svc:}"; then
+        bashio::log.info "Removing stale service host configuration for ${service_name}"
+        /opt/tailscale serve drain "${service_name}" >/dev/null 2>&1 || true
+        /opt/tailscale serve clear "${service_name}" >/dev/null 2>&1 || true
+    fi
+done
+
+# Configure and advertise each service
+for service in "${configured_services[@]}"; do
+    name=$(jq -re '.name' <<< "${service}")
```
lmagyar: bashio:jq ?

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+    name="${name#svc:}"
+    service_name="svc:${name}"
+    protocol=$(jq -re '.protocol' <<< "${service}")
+    port=$(jq -re '.port' <<< "${service}")
+    target=$(jq -re '.target' <<< "${service}")
+    path=$(jq -re '.path // empty' <<< "${service}")
+
+    options=()
+    case "${protocol}" in
```
lmagyar: protocol's value is restricted by the config schema, validate only that it is in the accepted list, exit.nok if not (fail during development), then use it directly in a single line

tailscale/DOCS.md
apfohl: I had the `svc:` prefix there in the first place, but after configuring it multiple times during testing I felt it annoying. But for parity to the `advertise_tags` I can revert this change. But honestly I would like the tags also more without the prefix. But that's just a personal preference. If it makes it more clean I'll revert. Just say. ;)

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
lmagyar: if ! "${protocol}" = @(http|https|tcp|tls-terminated-tcp); then
  exit.nok "Invalid value '${protocol}' for services.protocol"
fi

tailscale/rootfs/etc/s6-overlay/scripts/stage2_hook.sh
apfohl: Good point.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
apfohl: It was ment as a safe guard agains when Tailscale should decide to put other things in the .service section then actual services. But you are questioning this correctly. Why should they do it. I'll remove that guard.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+            ;;
+        tcp)
+            options+=("--tcp=${port}")
+            ;;
+        tls-terminated-tcp)
+            options+=("--tls-terminated-tcp=${port}")
+            ;;
+    esac
+
+    if [[ -n "${path}" ]]; then
```
lmagyar: bashio::var.has_value

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
apfohl: Ah, okay that's useful. I haven't caught this.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+        options+=("--set-path=${path}")
+    fi
+
+    # Clean any existing configuration for this service before reconfiguring
+    /opt/tailscale serve drain "${service_name}" >/dev/null 2>&1 || true
```
lmagyar: at least a warning should be logged when removing, otherwise when the serve below fails during recreation, nobody will know the root cause

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
apfohl: That will be not necessary anymore, when we make `svc:` mandatory.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+    /opt/tailscale serve clear "${service_name}" >/dev/null 2>&1 || true
+
+    bashio::log.info "Advertising service host for ${service_name}"
+    if ! /opt/tailscale serve --service="${service_name}" --yes "${options[@]}" "${target}"; then
```
lmagyar: I would add "serve" and "target" to the options array above, this is how it is implemented everywhere else
do we need --yes?

tailscale/rootfs/etc/s6-overlay/scripts/stage2_hook.sh
apfohl: True!

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
apfohl: Didn't know that.

tailscale/DOCS.md
lmagyar: Agree, we should remove tag: also, in a separate PR. If this PR is accepted, I will make a tag: removing PR. :)

apfohl: I would keep it in both places here and do another PR removing them together. So there is no mixed state in between.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
lmagyar: I mean `name="${name#svc:}"` currently it is not needed, there is no svc: in `services.name`.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
apfohl: I'm not very good in bash. Thanks for the advice.

tailscale/DOCS.md
lmagyar: It is your decision. :)

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
apfohl: No it's not anymore. Earlier `tailscale serve` used to ask interactively when invoking with a service. I tried manually. It's no longer the case. I'll remove it.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
lmagyar: Another idea. If https or tls-terminated-tcp is used, shouldn't we check once (!!!) that https support is enabled? Like in 
https://github.com/hassio-addons/app-tailscale/blob/main/tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-homeassistant/run#L24

Though we can say, that share_homeassistant is for the less experienced users, with more checks. :D

lmagyar:
And now a general comment.

The share_homeassistant option worked similarly a few years ago ("recreating" TS serve/funnel config), and as far as I remember there were complaints, that it comflicted manually configured serve configs, that's why that was changed to a front running serve/funnel with --bg=false.

So, when this change is merged, it will delete previously manually configured serve configs for the services. Hmmm, I think we should mark it (potentially) breaking change. The removal is logged, so I think it's fine.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+    fi
+done
+
+# Configure and advertise each service
+for service in "${configured_services[@]}"; do
+    name=$(jq -re '.name' <<< "${service}")
+    name="${name#svc:}"
+    service_name="svc:${name}"
+    protocol=$(jq -re '.protocol' <<< "${service}")
+    port=$(jq -re '.port' <<< "${service}")
```
lmagyar: My last question: can we use the same eg. 443 port here, as the share_homeassistant service uses? Services are running on a virtual tailnet name/IP, so it shouldn't be a problem. I'm asking, because I'm lazy to test it myself. :D

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
lmagyar: Don't they use it to check if https is enabled and ask whether they should enable it if not? AFAIK they don't do it in a non-interactive cli and just emit an error, but I'm not 100% sure about this. :(

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,117 @@
...
+
+# Read configured services
+# Note: bashio::config emits one JSON object per line for array options
+readarray -t configured_services < <(bashio::config 'services')
+
+# Build a list of configured service names (without the svc: prefix)
+for service in "${configured_services[@]}"; do
+    name=$(jq -re '.name' <<< "${service}")
+    name="${name#svc:}"
+    configured_names+=("${name}")
```
lmagyar: One more thing, to get configured_names, you can use bashio::config 'services.names' (if I remember correctly), and add some jq or grep to delete the svc: prefixes if needed, so this whole loop can be a one-liner I think.

apfohl:
@lmagyar Thanks for all the good feedback. I will work it in over the next week.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
apfohl: Yes that should be possible. I will test this once i have implemented the changes.

lmagyar:
I've just marked it draft until you work on it.

apfohl:
@lmagyar I implemented most of it now. But it feels a hard to keep track. Is it okay if I close the worked of conversations? Or do you want to close them during a review later? I still need to test all the variants to expose the services now.

lmagyar:
Feel free to do as you wish. :) I will remember what I've commented.

Meanwhile I'm thinking to leave share_homeassistant option as-is: that can use funnel also, but funnel is incompatible with services, and the current foreground running service behind share_homeassistant doesn't interfere with services (the port thing).

Maybe we can add a comment to the share_homeassistant option to the docs, that if somebody wants to share HA with services, use the services option instead.

apfohl:
Tried my best, to address everything in a meaningful way. I need to still test all protocols though.

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
apfohl: I tried using 443 on multiple services. It works as expected. 👍

apfohl:
Alright, I created a test [Home Assistant App](https://codeberg.org/apfohl/home-assistant-apps/src/branch/main/protocol-test) providing HTTP and TCP ports and configured the corresponding services in my Tailnet:

```yaml
services:
  - name: svc:http
    target: http://127.0.0.1:8090
    protocol: http
    port: 8090
  - name: svc:https
    target: http://127.0.0.1:8090
    protocol: https
    port: 443
  - protocol: tls-terminated-tcp
    name: svc:tls-tcp
    target: tcp://127.0.0.1:8092
    port: 8093
  - name: svc:tcp
    target: tcp://127.0.0.1:8092
    protocol: tcp
    port: 8092
```

And a [client](https://codeberg.org/apfohl/home-assistant-apps/src/branch/main/protocol-test/client):

```shell
go run . --tailnet=tailaa.ts.net --http-service=http --http-port=8090 --https-service=https --tcp-service=tcp --tcp-port=8092 --tls-tcp-service=tls-tcp --tls-tcp-port=8093

ADDRESS                              PROTOCOL   RESULT DETAIL
http.tailaa.ts.net:8090              HTTP       PASS
https.tailaa.ts.net:443              HTTPS      PASS
tcp.tailaa.ts.net:8092               TCP        PASS
tls-tcp.tailaa.ts.net:8093           TLS-TCP    PASS
```

tailscale/translations/en.yaml
```diff
@@ -74,6 +74,14 @@ configuration:
       internet.
       Only ports 443, 8443, and 10000 are allowed by Tailscale.
       Port 443 is used by default.
+  services:
+    name: Service hosts
+    description: >-
+      This option allows you to advertise other local services running on this
+      device as Tailscale Services. Each service needs a name, a local target
+      address, a protocol, and a port to expose it on. The service name must
+      include the `svc:` prefix.
```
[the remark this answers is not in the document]
apfohl: fixed the PR description.

lmagyar:
Give me some time... :)

tailscale/rootfs/etc/s6-overlay/s6-rc.d/share-services/run
```diff
@@ -0,0 +1,103 @@
...
+# Remove services that are no longer configured in this app
+for service_name in "${current_names[@]}"; do
+    if ! printf '%s\n' "${configured_names[@]}" | grep -qx "${service_name}"; then
+        bashio::log.info "Removing stale service host configuration for ${service_name}"
+        if ! /opt/tailscale serve drain "${service_name}" >/dev/null 2>&1; then
+            bashio::log.warning "Failed to drain stale service host ${service_name}"
+        fi
+        if ! /opt/tailscale serve clear "${service_name}" >/dev/null 2>&1; then
+            bashio::log.warning "Failed to clear stale service host ${service_name}"
+        fi
```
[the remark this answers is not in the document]
lmagyar: On draining: it's not an issue, the whole app is just started, there won't be any connections, if there are, they will be aborted, I'm fine with it.

On cleanup: yes, that is true, fixed it, push will follow soon.

lmagyar:
@coderabbitai review

lmagyar:
I've made some changes. :) I really liked your PR! Could you please test the current state on your system?

But first, you've found 3 bugs in other codes:
- I forget to add `shopt -s extglob` to another PR (#673), because that is enabled in my fork by the bash-completion package, but that is not used in the official app, ooops (OK, finally I removed it here, see below).
- I didn't notice, that bashio::exit.nok has only 1 argument for the message, not like the bashio::log functions, created a PR in the bashio repo (https://github.com/hassio-addons/bashio/pull/290).
- Due to misconfiguration I've run into the "S6 restarts a failing service during container startup forever" issue, I had to consult the S6 lib developer and finally created a not too small PR #729.
  
Changes I've made:
- code
  - I read the current config with `/opt/tailscale serve status --json`, because it has protocol info, while this is missing from`/opt/tailscale serve get-config --all` (`/opt/tailscale serve status --json` is basically the dump of the internal config from `curl -sf --unix-socket /var/run/tailscale/tailscaled.sock http://local-tailscaled.sock/localapi/v0/serve-config`) and I moved it to the beginning.
  - Moved the HTTPS and protocol value validation to the beginning, and using jq, not extglob checks inside the loop.
  - Added `sort -u` to the configured_names and current_names list, because 1 service can have multiple endpoints (eg. port 80 and 443), plus each endpoint can have multiple pathes.
  - I drain/clear "current_names \ configured_names" (stale services) and "current_names ⋂ configured_names" (changing/will be recreated services) before the loop. Otherwise if a service have 2 endpoint and we drain/clear it inside the loop, only the second endpoint will be configured at the end. This was mentioned in the AI review also.
  - Renamed the S6 service from share-services to only services (like taildrive and other services)
  - On first startup it imports the currently configured services once! Logs what it can't import (eg. target = text:"foo bar"), and adds only what is not already present in the yaml config. There is some jq black magic.
  - share_homeassistant depends on this service (waits for it), because otherwise I run into some race condition issues: "Another client is changing the serve config; please try again. \ sending serve config: Preconditions failed: etag mismatch"
- docs
  - Moved the example config to the beginning of the docs. That is not the default, but an example config also.
  - I used `advertised_tags` insted of `tags` PR #585 will rename it.
  - Some minimal typo, formatting changes.
  - Removed the warning to manually "import" existing config, I wrote the script for it (see above).
  - **I still can't figure out why only the "Service host" name is used from the translation file, but nothing else...** :(
 
I plan to release it on my fork, let see how the automatic import works in real life. Before we screw up 60k+ configs. :)

tailscale/rootfs/etc/s6-overlay/s6-rc.d/services/run
```diff
@@ -0,0 +1,224 @@
...
+  fi
+
+  # Filtering for importable services
+  if importable_services=$(bashio::jq "${current_config}" '
+    .Services // {} | to_entries[] | .key as $name
+    | .value.Web | to_entries[] | (.key | split(":")[1]) as $port
+    | .value.Handlers | to_entries[] | .key as $path
+    | .value | to_entries[] | select(.key=="Proxy")
+    | { "name": $name, "port": $port, "path": $path, "target": .value }' \
+    -e)
```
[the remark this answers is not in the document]
lmagyar: Fine, I've tested only http and https... Now I will test tcp and TLS-terminated-TCP entries also... Never ending story...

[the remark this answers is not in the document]
lmagyar: fixed in the new commits

lmagyar:
@coderabbitai review

apfohl:
Thank you!!! I will test it next week. That stirred the pot a bit... Great you like it. I'm using my current state since the beginning already. I will test it thoroughly.

lmagyar:
Finally I've released it on my fork let's see how it works. Fingers crossed.

lmagyar (approved):
Now it is LGTM.

lmagyar:
At the moment ~150 reported installs are running with my latest version with this PR merged in. No complains up to now, fingers crossed. :)

frenck (approved):
Thanks, @apfohl 👍 

../Frenck  

<p>
  <a href="https://www.linkedin.com/in/frenck/"></a>
  &nbsp;&nbsp;
  <a href="https://youtube.com/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://bsky.app/profile/frenck.social"></a>
  &nbsp;&nbsp;
  <a href="https://fosstodon.org/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.instagram.com/frenck/"></a>
  &nbsp;&nbsp;
  <a href="https://www.threads.net/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.facebook.com/frenck.dev/"></a>
  &nbsp;&nbsp;
  <a href="https://x.com/frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.tiktok.com/@frenck.nl"></a>
</p>

Blogging my personal ramblings at <a href="https://frenck.dev">frenck.dev</a>

apfohl:
Sorry, haven't found any time to test this again. I'm a bit surprised about the sudden merge. Don't get me wrong, I appreciate that it's going live now. Thx.

lmagyar:
Nearly 500 people installed this via my fork, no complains, but fingers crossed. :)

tailscale/config.yaml
```diff
@@ -79,6 +80,12 @@ schema:
   login_server: url
   share_homeassistant: list(disabled|serve|funnel)
   share_on_port: match(^(?:443|8443|10000)$)
+  services:
+    - name: "match(^svc:[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$)"
+      target: "match(^(?:https?|tcp)://127\\.0\\.0\\.1:([1-9]\\d{0,3}|[1-5]\\d{4}|6[0-4]\\d{3}|65[0-4]\\d{2}|655[0-2]\\d|6553[0-5])$)"
```
loongyh, line 85: Pls add support for `https+insecure://` as services with self-signed certs (such as the unifi network application addon) does not work.
`^(?:https?|https\+insecure|tcp)://127\\.0\\.0\\.1:([1-9]\\d{0,3}|[1-5]\\d{4}|6[0-4]\\d{3}|65[0-4]\\d{2}|655[0-2]\\d|6553[0-5])$`

lmagyar: Nice catch. The import already tried to import it. I've added docs, translation and proper escaping in the config regex (`\\+`), no code changes. Already testing it.

lmagyar: And it's released.

loongyh: Awesome, just tested it works. Thanks for working on this!

lmagyar: > just tested it works

Thanks for the feedback!

